PR #113, opened 2026-01-07
fix: split messages before formatting to preserve code blocks

anneschuth:
## Summary

- Fixed code blocks being incorrectly split across messages
- Reordered `flush()` to split on raw markdown BEFORE formatting
- Added safety check to verify ``` markers are actually closing markers

## Problem

Code blocks were being incorrectly split in messages. When content exceeded the soft break threshold:
1. The formatter ran first, converting `\`\`\`diff` to `\`\`\`` (Slack strips language hints)
2. The splitter then couldn't distinguish opening from closing ``` markers
3. Opening markers were incorrectly treated as "code_block_end" breakpoints

## Solution

1. **Reorder operations**: Split content on raw markdown (where `\`\`\`language` patterns are intact), then format each chunk
2. **Add safety check**: In `findLogicalBreakpoint`, verify we're inside a code block before treating ``` as a closing marker

## Test plan

- [x] All 1041 existing tests pass
- [x] Added 3 new tests for code block splitting edge cases
- [x] Build succeeds
- [x] Type check passes
- [x] Lint passes (only pre-existing warnings)

### DIFF
--- a/src/session/streaming.test.ts
+++ b/src/session/streaming.test.ts
@@ -763,6 +763,60 @@ describe('findLogicalBreakpoint with code blocks', () => {
     const result = findLogicalBreakpoint(content, 0);
     expect(result?.type).toBe('code_block_end');
   });
+
+  test('does not break at opening code block marker (bug fix)', () => {
+    // This was a bug: when content had a code block starting in the search window,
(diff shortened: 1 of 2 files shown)


anneschuth:
Closing due to merge conflicts with main. The key fix (split before format) will be re-implemented in a future PR.

PR #325, opened 2026-04-17
fix: surface skipped-file warnings on session start + recognize .har/.log as text

shaders:
## Summary

  Two related fixes for file attachment handling:

  1. **Skipped files on session start were silently dropped.** When a user started
     a new session by @-mentioning the bot with attachments, any unsupported file
     was stripped from the message without feedback. The `⚠️ Some files could not
     be processed` warning was only posted for *follow-up* messages via
     `MessageManager.handleUserMessage`. The initial-message path in
     `startSession` called `buildMessageContent()` which threw the `skipped` array
     away.

  2. **`.har` and `.log` files were classified as unsupported.** Mattermost and
     Slack report these with a generic MIME type (`application/octet-stream`),
     and neither extension was in `TEXT_FILE_EXTENSIONS`, so they fell through
     to `categorizeFile() → 'unsupported'` and were dropped.

  ## Changes

  - `src/operations/streaming/handler.ts`
    - Add `.har` and `.log` to `TEXT_FILE_EXTENSIONS`.
    - Extract `formatSkippedFilesFeedback()` as an exported helper (was a
      private duplicate inside `MessageManager`).
  - `src/operations/message-manager.ts`
    - Use the shared `formatSkippedFilesFeedback()`; remove the duplicate.
  - `src/session/lifecycle.ts`
    - After `claude.sendMessage(content)` in `startSession`, call
      `processFiles()` on the attached files and post the feedback message if
      anything was skipped — mirroring the follow-up path in `MessageManager`.

  ## Test plan

  - [x] `bun run typecheck` — passes
  - [x] `bun test src/` — 1912 / 1912 passing
  - [x] `bun run build` — bundles cleanly
  - [x] Manual: start a session with a `.har` + image attachment; verify the
        HAR is inlined as a text block and the image is sent as a vision block.
  - [x] Manual: start a session with an unsupported file (e.g. `.xlsx`); verify
        the bot posts the `⚠️ Some files could not be processed` warning in
        the thread.
  - [x] Manual: send a follow-up message with a skipped file in an existing
        session to confirm the existing warning flow still works.

  ## Notes

  - `processFiles()` is called twice on the session-start path (once inside
    `buildMessageContent`, once to collect `skipped`). This matches the
    existing pattern in `MessageManager.handleUserMessage` and keeps the
    change minimal. A future refactor could change `buildMessageContent()` to
    return `{ content, skipped }`, but that touches many call sites.
  - Worktree-prompt and context-prompt paths were not touched — they queue
    files for later processing; if needed, the same pattern can be applied
    where those queues are drained.


anneschuth (changes requested):
Thank you for this, Max — the bug is real, the diff is clean, and extracting `formatSkippedFilesFeedback` as a shared helper is a nice touch. Before merging I'd like to ask for two changes:

**1. Refactor `buildMessageContent` to return `{ content, skipped }` (which you already suggest in the PR notes).**

The current shape calls `processFiles` twice on the session-start path: once inside `buildMessageContent` at `lifecycle.ts:874`, and once in the new block at `lifecycle.ts:899`. That means duplicate `platform.downloadFile` calls, duplicate gzip decompression / zip extraction, and duplicate `log.warn("Skipped file …")` lines (`handler.ts:922`) per skipped file on every session start with attachments. Mirroring `handleUserMessage` is fair, but the existing pattern there has the same smell.

Threading `skipped` through the return value fixes the root cause once. As a bonus, it also closes the context-prompt gap you flagged: `handleContextDecision` in `context-prompt/handler.ts:419/463/488` currently drops `skipped` too, so a user who attaches an unsupported file and lands in the mid-thread context-prompt flow still gets silence. A single-contract refactor makes every call site surface the warning uniformly, with no extra work per call site.

The churn is mechanical — the `BuildMessageContentCallback` type, `session-context/types.ts:115`, the three `ctx.ops.buildMessageContent(…)` call sites in `lifecycle.ts`, the two in `worktree/handler.ts` and `context-prompt/handler.ts`, plus test doubles. Happy to help if anything gets sticky.

**2. A regression test in `lifecycle.test.ts`.**

Something that spawns `startSession` with an unsupported file and asserts `platform.createPost` is called with the `⚠️ Some files could not be processed` warning. The three manual cases in your test plan are exactly right; an automated one locks the behavior in and will catch the next refactor that drops the feedback path.

Neither is a huge lift, and both come from your own PR description — just want them in the diff rather than in a follow-up. Thanks again for the careful writeup!

shaders:
Thanks for the thorough review — both asks made the diff cleaner. Done:

  ## 1. `buildMessageContent` now returns `{ content, skipped }`

  Single contract, `processFiles` runs once, zero duplicate `platform.downloadFile` / decompression / `log.warn` on skipped files.

  Mechanical updates as you flagged, plus a couple of adjacent wins:

  - **Type flow:** `BuildMessageContentCallback`, `session-context/types.ts:115`, `worktree/handler.ts:389`,
  `context-prompt/handler.ts:374`, and all four test doubles now speak `BuiltMessageContent`. `ContentBlock` stays inside the payload
  where it belongs (the `claude.sendMessage()` boundary).
  - **Shared posting helper:** new `postSkippedFilesFeedback(platform, threadId, skipped)` in `streaming/handler.ts` — no-op on empty,
  posts the ⚠️ warning otherwise. Every call site is now one line.
  - **Context-prompt gap closed:** all three paths in `context-prompt/handler.ts` (`handleContextPromptTimeout`, `offerContextPrompt`
  zero-message branch, `offerContextPrompt` single-message auto-include) now surface skipped files. The mid-thread-attachment silence
  you called out is gone.
  - **Worktree path:** mid-session worktree replay in `worktree/handler.ts:671` now surfaces skipped files too.
  - **Dead code removed:** the `buildMessageContent` call at `lifecycle.ts:1201` was building content only to recompute `typeof content
   === 'string' ? content : message`, which is identically `message` in both branches. Dropped the call and pass `message` straight
  through to `offerContextPrompt`, which handles its own files.
  - **`MessageManager.handleUserMessage`:** the local `processFiles` call is gone; it destructures the callback result instead.

  ## 2. Regression test in `lifecycle.test.ts`

  Added two tests under `describe('startSession skipped-file feedback', ...)`:

  - Spawns `startSession` with an unsupported `.xlsx`, mocks `ctx.ops.buildMessageContent` to return a populated `skipped` array,
  asserts `platform.createPost` was called with a body containing `⚠️`, `Some files could not be processed`, `report.xlsx`,
  `Unsupported file type`, and the suggestion text.
  - Negative case: empty `skipped` ⇒ no warning post.

  The existing comment in this file noted that testing `startSession` directly requires mocking Claude CLI spawn; I used the same
  `mock.module('../claude/cli.js', ...)` pattern that `worktree/handler.test.ts` already uses, plus a
  `mock.module('../claude/quick-query.js', ...)` so the fire-and-forget metadata/tag suggestions don't try to call a real Claude — and
  deliberately at the deeper `quick-query` layer so the suggestion tests' own module mocks aren't clobbered in the same run.

  **RED-GREEN verified:** commented out `postSkippedFilesFeedback` at the session-start site → test fails on
  `expect(warning).toBeDefined()`. Restored → passes. Also added two unit tests for `postSkippedFilesFeedback` itself in
  `tests/unit/file-attachments.test.ts` (empty ⇒ no-op, non-empty ⇒ formatted warning with names/reasons/suggestions).

  ## Results

  - `bun run typecheck` — clean
  - `bun test` — **2034 pass / 0 fail** (173 skip = Docker integration)
  - `bun run build` — bundles fine

  Diff: 11 files, +263 / −82. Ready for another look.

anneschuth (changes requested):
Max, one last thing — the CI failure is real but subtle. Thanks for the patience.

**Diagnosis:** Your `MockClaudeCli` in `lifecycle.test.ts` only has 6 methods (`isRunning`, `kill`, `start`, `sendMessage`, `on`, `interrupt`), but Bun's `mock.module()` is process-global in **Bun 1.3.13** (CI version). The mock leaks into `src/claude/cli.test.ts`, where the real `ClaudeCli` tests then fail with `cli.getStatusFilePath is not a function`, `cli.getLastStderr is not a function`, etc. — the real class has a wider surface area than your mock.

There's a pre-existing identical mock in `worktree/handler.test.ts` that doesn't trigger this, presumably because of Bun's file-discovery ordering — but the new mock in `session/` comes before `src/claude/` alphabetically, so it poisons the cache by the time `cli.test.ts` loads.

**Fix:** make the mock expose every method `cli.test.ts` probes, with return values that also satisfy the "not started" assumption that suite relies on (`isRunning() === false`, `interrupt() === false`, etc.). Verified locally on Bun 1.3.13 with `bun test src/ --coverage` — 1914/1914 pass.

Suggestion below. `isRunning() { return false }` (was `true`) is the only semantic change; your new `startSession` tests don't depend on `true` because the `claude.sendMessage(content)` call isn't gated by `isRunning()` on the initial-send path.

src/session/lifecycle.test.ts
```diff
@@ -1,9 +1,31 @@
...
+mock.module('../claude/cli.js', () => ({
+  ClaudeCli: class MockClaudeCli {
+    isRunning() { return true; }
+    kill() { return Promise.resolve(); }
+    start() {}
+    sendMessage() {}
+    on() {}
+    interrupt() {}
+  },
+}));
```
anneschuth, line 62: Expanding `MockClaudeCli` to cover the full `ClaudeCli` surface so Bun 1.3.13's process-global module mock doesn't break `src/claude/cli.test.ts`:

```suggestion
// Mock ClaudeCli so startSession doesn't spawn a real Claude process.
// Must be declared before importing lifecycle so the module cache picks it up.
// NOTE: Bun's mock.module() persists globally across the test run in Bun
// 1.3.13+, so this mock can leak into src/claude/cli.test.ts. The mock class
// must expose the same method surface as the real ClaudeCli, with return
// values that also satisfy cli.test.ts's "freshly constructed, not started"
// assumptions.
mock.module('../claude/cli.js', () => ({
  ClaudeCli: class MockClaudeCli {
    isRunning() { return false; }
    kill() { return Promise.resolve(); }
    start() {}
    sendMessage() {}
    sendToolResult() {}
    on() {}
    off() {}
    interrupt() { return false; }
    getStatusFilePath() { return null; }
    getStatusData() { return null; }
    getLastStderr() { return ''; }
    isPermanentFailure() { return false; }
    getPermanentFailureReason() { return null; }
    startStatusWatch() {}
    stopStatusWatch() {}
  },
}));
```

anneschuth (changes requested):
See inline suggestion on lines 3-14 for the Bun 1.3.13 mock-leak fix. Happy to help if anything else comes up — the rest of the PR is in great shape.

shaders:
Thanks for the diagnosis — pushed `d80e2e8`. Expanded `MockClaudeCli` to match the real class's method surface, flipped `isRunning()`
   / `interrupt()` to return `false` so `cli.test.ts`'s "not started" assumptions hold. Added a comment on the mock explaining why, so future edits don't reintroduce the leak.

shaders:
let me check why this is still failing.

shaders:
Pushed `b1471d6`. The earlier expansion fixed the method surface but missed the behavioral contracts that cli.test.ts actually asserts against — debug flag from env, `sendMessage`/`sendToolResult` throwing when not running, `start()` validating `platformConfig`. Since `mock.module()` leaks into cli.test.ts, those tests run through the mock and need the same semantics as the  real class.

  Added a constructor (reads `DEBUG`/`--debug`), a `running` flag flipped by `start()`/`kill()`, and the pre-spawn validation. The  session tests here still work because `startSession` calls `start()` before `sendMessage` — `running` is true by then, so no spurious  throws.

shaders:
let me check that one too

shaders:
Can we try re-runing?

 Unrelated flake in the tool_use integration test — this one has a history (see #320, previously skipped + revived). It timed out at 30840ms just past the 30s wait, not a logical failure, and nothing in this diff touches the tool_use / permission-prompt /  Mattermost-post code paths. Re-running the job should go green.

Happy to bump the timeout or add a retry in a separate PR if it keeps flaking.

anneschuth:
Green enough to merge once the mattermost integration flake clears (rerunning now — it's been flaky all week, nothing to do with your change). Thanks for sticking with four rounds of this — the PR is in great shape. Once mattermost flips green I'll squash-merge.

anneschuth (approved):
All checks green. Nicely done — thanks for the careful work.

PR #328, opened 2026-04-21
feat: opt-in multi-account Claude support with rate-limit aware round-robin

shaders:
## Summary

  Adds an **opt-in** pool of Claude accounts so sessions round-robin across
  multiple subscriptions / API keys instead of sharing one token budget. When an
  account hits a rate limit the bot puts it in cooldown and routes new sessions
  to another account; the cooling account resumes automatically when the
  extracted reset time passes.

  Omitting `claudeAccounts` from YAML leaves the bot in single-account mode —
  every session inherits `process.env` exactly as before. Zero impact on
  existing installs.

  ### Config (opt-in)

  ```yaml
  claudeAccounts:
    # OAuth (prepared via `HOME=<path> claude login` against that home)
    - id: primary
      home: /home/bot/.claude-accounts/primary
    - id: backup
      displayName: Backup (Pro)
      home: /home/bot/.claude-accounts/backup

    # API-key billed (no HOME override needed)
    - id: shared-api
      apiKey: sk-ant-api03-xxxxxxxx...

  How it works

  - Spawn-time env override. ClaudeCli accepts an account option and
  sets either HOME/USERPROFILE (OAuth → reads .credentials.json,
  .claude/projects/*, MCP config from that dir) or ANTHROPIC_API_KEY
  (API-billed, HOME untouched). Conflicting env vars from the parent
  (ANTHROPIC_API_KEY under OAuth mode; CLAUDE_CODE_OAUTH_TOKEN in either
  mode) are cleared so the selected account actually wins.
  - Pooling. AccountPool does round-robin acquire(), accepts a
  preferredId (used by resume so OAuth history under a specific HOME is
  recovered), and tracks cooldown per account. Session exits release the slot
  via a one-place helper that covers both the happy path and all failure
  branches.
  - Persistence. PersistedSession.claudeAccountId remembers which account
  a session ran under. Resume passes it as preferredId. Legacy sessions
  without the field fall back to default env — no migration needed.
  - Rate-limit detection. Scans stderr and error-flavored result events
  (gated on subtype.startsWith('error') or is_error: true so assistant
  text that happens to mention "rate_limit_error" can't cool the account) for
  phrases like usage limit reached, rate_limit_error, 429 … rate limit,
  quota exceeded. Extracts the reset time from retry after N seconds,
  Resets in N units, JSON unix timestamps, or HH:MM UTC. Default fallback:
  1 hour. When detected, the pool marks the account cooling and posts a
  warning in the thread.
  - UI. Session header: 🔑 Claude account: <displayName> when
  multi-account mode is active. Channel sticky: 🔑 N accounts or
  🔑 A/N accounts (K cooling).

  Preserving account on session restart

  !cd and !permissions interactive respawn Claude. The restart helper now
  re-embeds the current account into fresh ClaudeCliOptions and rebinds the
  'rate-limit' listener along with 'event' / 'exit' so a mid-session
  rate-limit after a restart still enters cooldown.

  Files touched

  New

  - src/claude/account-pool.ts — round-robin + cooldown (+ 17 unit tests)
  - src/claude/rate-limit-detector.ts — pure parser (+ 19 unit tests)
  - src/operations/commands/restart-rebind.test.ts — regression lock on the three-listener rebind

  Modified

  - src/config/types.ts — ClaudeAccount type + Config.claudeAccounts?
  - src/claude/cli.ts — ClaudeCliOptions.account, buildChildEnv, 'rate-limit' event, isErrorResultEvent gate
  - src/session/{manager,lifecycle,types}.ts — pool wiring, claudeAccountId on Session, handleRateLimit, releaseAccountIfHeld on all
  exit paths
  - src/persistence/session-store.ts — claudeAccountId?: string
  - src/operations/commands/handler.ts — account preserved on restart, 🔑 row in header
  - src/operations/sticky-message/handler.ts — pool summary in status bar
  - src/operations/session-context/types.ts — 5 new pool ops
  - CLAUDE.md, CHANGELOG.md — docs

  Verified

  - bun run typecheck — clean
  - bun test — 1954/0 (prior baseline: 1912)
  - bun run build — clean bundle
  - Manual E2E with three accounts on an isolated Mattermost test channel:
    - Round-robin cycles through all accounts.
    - Each account's session writes its JSONL transcript into its own
  $HOME/.claude/projects/...; nothing lands in the bot operator's main
  ~/.claude.
    - claudeAccountId present on every persisted session; resumes pick the
  same account.
    - API-key account spawns correctly (verified via stderr showing no
  OAuth-related activity).

  Known follow-ups (separate PR)

  - Worktree mid-session respawn in src/operations/worktree/handler.ts:485,634
  rebuilds ClaudeCliOptions inline and currently doesn't thread the
  account or the 'rate-limit' listener — same class of bug as the !cd
  path that this PR fixes. Deliberately left alone here to keep the diff
  focused; will follow up with a small refactor that extracts the listener
  binding into a single helper and re-embeds the account via the existing
  sessionAccountOption pattern.
  - Config validation: reject accounts with both home and apiKey set (today
  the pool takes home wins silently).

  Test plan for reviewer

  - Remove claudeAccounts from config → bot behaves identically to pre-PR.
  - Configure two accounts → start two sessions in different threads → verify round-robin via logs / session header.
  - Restart bot with active sessions → they resume under the same account (check claudeAccountId in sessions.json pre- and
  post-restart).
  - Trigger or simulate a rate limit → verify the offending account is marked cooling, next session is assigned to a different account,
   and a warning posts to the thread.

### DIFF
--- a/src/claude/account-pool.ts
+++ b/src/claude/account-pool.ts
@@ -1,0 +1,148 @@
+/**
+ * AccountPool — round-robin selector over a pool of Claude accounts.
+ *
+ * Responsibilities:
+ * - Hand out an account for a new session (round-robin, skipping cooling accounts).
+ * - Track which accounts are currently in rate-limit cooldown so future sessions
+ *   route around them. Resume of existing sessions bypasses cooldown because the
+ *   conversation history lives under that account's HOME and can't be moved.
+ * - Track usage counts for UI display (sticky message).
+ *
+ * Single-account mode: pass an empty array (or `undefined`) to the constructor
+ * and every method returns `null` — the bot then falls back to `process.env` as
+ * it does today.
+ */
+import type { ClaudeAccount } from '../config/types.js';
+import { createLogger } from '../utils/logger.js';
+
+const log = createLogger('account-pool');
+
+/** Snapshot of pool state for UI/debug. */
+export interface AccountPoolStatus {
+  id: string;
+  displayName: string;
+  activeSessions: number;
+  coolingUntil: number | null; // epoch ms, null = available
+}
+
+export class AccountPool {
+  private readonly accounts: ClaudeAccount[];
+  private readonly byId: Map<string, ClaudeAccount>;
+  private readonly activeCounts: Map<string, number> = new Map();
+  private readonly coolingUntil: Map<string, number> = new Map();
+  private roundRobinIndex = 0;
+
+  constructor(accounts?: ClaudeAccount[]) {
+    this.accounts = (accounts ?? []).filter((acc) => {
+      const hasAuth = !!acc.home || !!acc.apiKey;
+      if (!hasAuth) {
+        log.warn(`Claude account ${acc.id} has neither home nor apiKey — ignoring`);
+      }
+      return hasAuth;
+    });
+    this.byId = new Map(this.accounts.map((acc) => [acc.id, acc]));
+    for (const acc of this.accounts) {
+      this.activeCounts.set(acc.id, 0);
+    }
+  }
+
+  /** True when no accounts are configured — caller should use default env. */
+  get isEmpty(): boolean {
+    return this.accounts.length === 0;
+  }
+
+  /** Number of configured accounts. */
+  get size(): number {
+    return this.accounts.length;
+  }
+
+  /**
+   * Acquire the next available account via round-robin, skipping cooling ones.
+   *
+   * If `preferredId` is given and that account exists, it is returned as-is —
+   * even if it's currently cooling. This path is used when resuming an existing
+   * session whose conversation history lives under that account's HOME.
+   *
+   * Returns `null` when the pool is empty, or when every account is cooling
+   * and no `preferredId` was supplied.
+   */
+  acquire(preferredId?: string): ClaudeAccount | null {
+    if (this.isEmpty) return null;
+
+    if (preferredId) {
+      const preferred = this.byId.get(preferredId);
+      if (preferred) {
+        this.incrementActive(preferred.id);
+        return preferred;
+      }
+      log.warn(`Preferred account "${preferredId}" not in pool — falling back to round-robin`);
+    }
+
+    const now = Date.now();
+    const n = this.accounts.length;
+    for (let i = 0; i < n; i++) {
+      const idx = (this.roundRobinIndex + i) % n;
+      const candidate = this.accounts[idx];
+      const cooling = this.coolingUntil.get(candidate.id) ?? 0;
+      if (cooling <= now) {
+        this.roundRobinIndex = (idx + 1) % n;
+        this.incrementActive(candidate.id);
+        return candidate;
+      }
+    }
+
+    log.warn(`All ${n} accounts are in rate-limit cooldown`);
+    return null;
+  }
+
+  /**
+   * Release an account — caller invokes this when a session ends so usage
+   * accounting stays accurate. No-op if the id isn't in the pool.
+   */
+  release(accountId: string): void {
+    const current = this.activeCounts.get(accountId);
+    if (current === undefined) return;
+    this.activeCounts.set(accountId, Math.max(0, current - 1));
+  }
+
+  /**
+   * Mark an account as rate-limited until `untilEpochMs`. Subsequent `acquire()`
+   * calls without `preferredId` will skip this account until the timestamp passes.
+   */
+  markCooling(accountId: string, untilEpochMs: number): void {
+    if (!this.byId.has(accountId)) {
+      log.warn(`markCooling called for unknown account "${accountId}"`);
+      return;
+    }
+    const existing = this.coolingUntil.get(accountId) ?? 0;
+    // Only extend cooldown, never shorten it.
+    if (untilEpochMs > existing) {
+      this.coolingUntil.set(accountId, untilEpochMs);
+      const minutes = Math.ceil((untilEpochMs - Date.now()) / 60000);
+      log.info(`Account "${accountId}" cooling for ~${minutes}min`);
+    }
+  }
+
+  /** Look up an account by id. Returns undefined for unknown ids. */
+  get(accountId: string): ClaudeAccount | undefined {
+    return this.byId.get(accountId);
+  }
+
+  /** Snapshot of pool state — for UI / sticky message / debug logs. */
+  status(): AccountPoolStatus[] {
+    const now = Date.now();
+    return this.accounts.map((acc) => {
+      const cooling = this.coolingUntil.get(acc.id) ?? 0;
+      return {
+        id: acc.id,
+        displayName: acc.displayName ?? acc.id,
+        activeSessions: this.activeCounts.get(acc.id) ?? 0,
+        coolingUntil: cooling > now ? cooling : null,
+      };
+    });
+  }
+
+  private incrementActive(accountId: string): void {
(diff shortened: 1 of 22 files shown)


shaders:
FYI, cannot be merged before https://github.com/anneschuth/claude-threads/pull/325 (used as base and need to pull latest cli.js fix after merge)

anneschuth:
@shaders thanks! could you rebase? #325 was merged

shaders:
rebased

anneschuth (commented):
Max — thank you for this. Serious piece of work, thoughtfully designed, well-tested. Before getting into the code, let me document the project stance since that shapes everything else.

## Product stance

claude-threads is optimized for **one bot = one Claude account**. The intended deployment pattern is: each user runs their own bot (or their team operator runs a dedicated bot per user) with their own token, in their own channel or with their own `@botname`. Rate limits are then a per-user problem, not a bot-level state machine, and the code stays small and easy to reason about.

Your PR targets a different (and legitimate) model: one bot serving a **pool** of accounts with round-robin dispatch and automatic cooldown. That's a real model — maybe the right one for some deploys — and your implementation is clean about it being fully opt-in.

So: this isn't the primary model I'm building toward, but I don't want to reject thoughtful, self-contained work that you're clearly a happy user of. The deciding factor for me: **does merging this hurt users who don't configure `claudeAccounts`?** I traced the hot paths carefully, and the answer is no. `AccountPool.acquire()` short-circuits on `isEmpty`, `buildChildEnv` returns `process.env` unchanged when `options.account` is undefined, the sticky-message pool row is gated on `!pool.isEmpty`, the session header `🔑` row only renders when an account is bound, persistence adds a single optional field, and the `'rate-limit'` listener is cheap to register with a no-op handler in single-account mode. The design is honest about being opt-in.

The cost lands on me as maintainer (new `SessionOperations` surface, `claudeAccountId` on `PersistedSession` going forward, every future `ClaudeCli` spawn site needing to remember the two invariants you call out). I'm OK taking that on in exchange for a feature that clearly works and a contributor who tests carefully.

## Overall impression

- **Design is sound.** Spawn-time env isolation via `HOME` / `ANTHROPIC_API_KEY` is the right primitive. Clearing the opposing env var (API key under OAuth mode, OAuth token under API-key mode) is the defensive thing to do. Good comment at `cli.ts:613-617` explaining why.
- **`isErrorResultEvent` gate on the detector is exactly right.** Scanning only error-flavored result events prevents the foot-gun where assistant text that explains "rate_limit_error" would cool the account. Your regression test makes the invariant explicit.
- **`acquireClaudeAccount(preferredId)` for resume is the right shape.** Preferring the original account even if cooling, because OAuth history lives under that HOME — correct and documented. `cli.ts:1044-1050`.
- **`releaseAccountIfHeld` is the right one-place helper.** Covers happy path + failure branches + cleanup. Guard against double-release via clearing `claudeAccountId` is a nice touch.
- **36 new unit tests on the pool and detector.** Coverage of the state transitions I'd want covered — cooling, re-acquire after expiry, preferred bypass, clamping release at zero, permissive matcher + strict caller gating.

Comments inline. Nothing blocks merge; the items below are notes for consideration or follow-up.

## Tests + CI

- `bun test` locally passes, 1954 tests.
- CI shows 6/7 green; the mattermost integration failure has been flaky all week on unrelated PRs — I'll rerun.
- No typecheck/lint issues observed.

## A small rebase

I merged #329 a few minutes ago, which removes the `startSession skipped-file feedback` block you anchored a hunk against in `lifecycle.test.ts`. Trivial rebase — just drop the removed block and reposition your new `describe` above `Session State Management`. Sorry for the churn.

Thanks again, Max.

src/claude/account-pool.ts
```diff
@@ -0,0 +1,148 @@
...
+export class AccountPool {
+  private readonly accounts: ClaudeAccount[];
+  private readonly byId: Map<string, ClaudeAccount>;
+  private readonly activeCounts: Map<string, number> = new Map();
+  private readonly coolingUntil: Map<string, number> = new Map();
+  private roundRobinIndex = 0;
+
+  constructor(accounts?: ClaudeAccount[]) {
+    this.accounts = (accounts ?? []).filter((acc) => {
+      const hasAuth = !!acc.home || !!acc.apiKey;
```
anneschuth, line 37: Small observation, not a blocker: the pool silently prefers `home` over `apiKey` when both are set (the `start()` branch at `cli.ts:610` picks `home` first). You've listed "reject accounts with both home and apiKey" as a known follow-up in the PR description — agreed. When that follow-up lands, I'd suggest doing the validation here at construction time (log + drop, same pattern as the `hasAuth` check), since it's the natural chokepoint and keeps `start()` branch-clean.

src/claude/rate-limit-detector.ts
```diff
@@ -0,0 +1,124 @@
...
+      second: 1000,
+      minute: 60_000,
+      hour: 3_600_000,
+      day: 86_400_000,
+    };
+    return now + value * unitMs[unit];
+  }
+
+  // JSON-ish: "reset_at": 1234567890  (seconds)
+  const unix = text.match(/["']?reset(?:_at)?["']?\s*[:=]\s*(\d{10,13})/);
```
anneschuth, line 102: Nit on the regex: `/["']?reset(?:_at)?["']?\s*[:=]\s*(\d{10,13})/` is permissive enough that it'll match `"preset": 1234567890000` (the `p` of `preset` doesn't anchor), or `reset_after=1234567890` (which is *not* an absolute reset time). Neither is likely in practice since the Anthropic API shape is `reset_at`, but given the detector already leans permissive on the phrase matchers and gets its safety from the caller-side `isErrorResultEvent` gate, a tighter anchor here — e.g. `\breset(?:_at)?\b` or requiring the surrounding `"` on both sides — would be cheap belt-and-suspenders.

Not a blocker; flagging because if a future Claude CLI version emits `reset_after` for a retry-after hint, this would silently misinterpret it as an absolute epoch.

src/claude/cli.ts
```diff
@@ -132,6 +168,7 @@ export class ClaudeCli extends EventEmitter {
   private statusFilePath: string | null = null;
   private lastStatusData: StatusLineData | null = null;
   private stderrBuffer = '';  // Capture stderr for error detection
+  private rateLimitEmitted = false;  // Fire 'rate-limit' only once per process
```
anneschuth, line 171: The `rateLimitEmitted` latch is per-`ClaudeCli`-instance and resets to `false` in `start()` (line 255). That's correct for the happy case, but worth flagging an edge:

A session that restarts via `!cd` or `!permissions interactive` gets a *new* `ClaudeCli` instance — so the latch resets, which is what you want (fresh process = fresh rate-limit state). Good.

But a session that stays on the same `ClaudeCli` instance and gets rate-limited a second time (e.g. first hit extracts a 10-minute reset, account cools, user somehow retries before the 10 minutes pass, second rate-limit fires with a 1-hour reset) will not re-emit. `AccountPool.markCooling` only extends cooldown (lines 117-123 of pool.ts), so the new 1-hour deadline wouldn't land. In practice this requires the session to survive a rate-limit hit and keep running, which I think it does today since `handleRateLimit` doesn't kill the session. Probably fine, but worth a comment noting the interaction, or — cheaper — make the latch re-open after some timeout so a fresh signal at least has a chance to extend the cooldown.

Not a blocker.

anneschuth (approved):
Rebase clean, all checks green. Thanks Max.

PR #330, opened 2026-04-21
follow-up review nits from #328 (multi-account)

shaders:
Follow-up to #328 (merged as `58c845c`). Three inline review nits from @anneschuth didn't make it into the merge — this PR lands them  on top of `main` as a small isolated change.

  All three were flagged as non-blockers on #328; grouping them together since they're all ~dozen-line fixes.

  ### 1. `AccountPool` rejects accounts with both `home` and `apiKey`

  The docs say these are mutually exclusive but the pool silently preferred `home` (via the `start()` branch in `cli.ts`). Moved validation into the `AccountPool` constructor filter — same log + drop pattern as the existing `hasAuth` check, so the spawn path stays branch-clean.

  File: `src/claude/account-pool.ts` · Test: `account-pool.test.ts`

  ### 2. Tighten `reset_at` epoch regex

  Previous pattern `/["']?reset(?:_at)?["']?\s*[:=]\s*(\d{10,13})/` would match `"preset": 1234567890` (no anchor on the `p`) or `reset_after=1234567890`. The second one is the real hazard — `reset_after` is a **relative** retry-after hint; parsing it as an absolute epoch would put cooldown decades into the future.

  Switched to `\breset(?:_at)?\b\s*["']?\s*[:=]\s*(\d{10,13})` and added regression tests for both cases.

  File: `src/claude/rate-limit-detector.ts` · Test: `rate-limit-detector.test.ts`

  ### 3. Rate-limit emit latch allows deadline extensions

  Replaced the boolean `rateLimitEmitted` with numeric `lastEmittedRateLimitDeadline`. A second rate-limit whose cooldown deadline advances by more than a minute now re-emits so `AccountPool.markCooling` (extend-only) can widen the window. Repeat hits at the same severity — including the stderr-chunk spam the boolean latch was designed to prevent — stay deduped. The 1-minute floor keeps `Date.now()`-based drift on relative hints (`"Resets in 10 minutes"`) from counting as a new event.

  File: `src/claude/cli.ts` · Test: `cli.test.ts` (dedupe / extend / earlier-deadline / non-rate-limit-text)

  ## Verified

  - `bun run typecheck` clean
  - `bun test src/` — **1959/1959** (+5 new tests)
  - `bun run build` clean
  - RED-GREEN verified on (2) and (3): both tests fail when the fix is reverted

  Thanks again for the review — you were right on all three.

### DIFF
--- a/src/claude/account-pool.test.ts
+++ b/src/claude/account-pool.test.ts
@@ -30,6 +30,22 @@ describe('AccountPool', () => {
       expect(pool.get('valid')).toBeDefined();
       expect(pool.get('api')).toBeDefined();
     });
+
+    it('drops accounts that have BOTH home and apiKey (mutually exclusive)', () => {
+      // home/apiKey are documented as mutually exclusive: `home` routes via
+      // OAuth, `apiKey` via API billing. Silently preferring one (as the old
+      // behavior did) hides misconfiguration; the pool should reject the
+      // account outright so the operator notices.
+      const pool = new AccountPool([
+        { id: 'oauth', home: '/tmp/a' },
+        { id: 'dual', home: '/tmp/b', apiKey: 'sk-ant-xxx' }, // invalid
+        { id: 'api', apiKey: 'sk-ant-yyy' },
+      ]);
+      expect(pool.size).toBe(2);
+      expect(pool.get('dual')).toBeUndefined();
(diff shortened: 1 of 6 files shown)


shaders:
Thanks for a great project!

anneschuth (approved):
All three fixes on-point; tests prove each works. Thanks Max.

PR #350, opened 2026-04-28
fix: sticky-by-thread Claude account binding (multi-account)

shaders:
## Problem

  In multi-account mode (`claudeAccounts:`), `AccountPool.acquire()` round-robins over accounts on session start. The chosen account's
  `$HOME` determines where the Claude CLI writes its conversation history (`<home>/.claude/projects/...`). The same id is also written
  to `sessions.json` so resume after a bot restart re-spawns Claude under the matching `$HOME`.

  These two writes happen on different code paths and aren't atomic together. In production, `sessions.json` ends up recording a
  different `claudeAccountId` than the `$HOME` Claude was actually spawned under. After a bot restart, resume picks the wrong `$HOME`,
  fails to find the JSONL, raises `Detected permanent failure: The conversation history for this session no longer exists`, and
  soft-deletes the session — users see `❌ Session cannot be resumed`.

  Real-world incident: 14 sessions soft-deleted in one restart cycle — 3 were genuine victims of this drift, the other 11 were normal
  `cleanStale`. The exact write-ordering that produces the drift was not localized; multiple paths can race on `claudeAccountId`. This
  PR removes the possibility of drift entirely rather than serializing the writes.

  ## Fix

  Sticky-by-thread account binding. `AccountPool.acquire(preferredId, threadId)` now picks `accounts[hash(threadId) % n]` when given a
  threadId. The `$HOME` chosen at spawn time and the `claudeAccountId` persisted to `sessions.json` are both derived from the same hash
   — they cannot drift apart under concurrent acquisitions.

  Selection priority:
  1. **preferredId** — honored even if cooling (resume invariant: OAuth history lives under that account's HOME and can't move)
  2. **threadId-sticky bucket** — round-robin fallback only if the sticky account is in rate-limit cooldown
  3. **round-robin** — unchanged

  Sticky-path acquires do **not** advance `roundRobinIndex`, so heavy use of one sticky thread does not silently shift how anonymous
  acquires distribute.

  Hash: FNV-1a 32-bit, pure, dependency-free.

  | File | Change |
  |---|---|
  | `src/claude/account-pool.ts` | sticky branch + `hashThreadId` helper |
  | `src/operations/session-context/types.ts` | extended ops signature `acquireClaudeAccount(preferredId?, threadId?)` |
  | `src/session/manager.ts` | pass threadId through to pool |
  | `src/session/lifecycle.ts` | `actualThreadId` on start, `state.threadId` on resume |
  | `src/claude/account-pool.test.ts` | 5 new tests (3 RED-GREEN regression defenders + 2 contract anchors) |

  ## Backward compatibility

  Pure addition. Existing `acquire(preferredId?)` callers continue to work. Resume path now also passes `state.threadId` so legacy
  sessions persisted before this fix re-derive the same sticky account when `state.claudeAccountId` is missing.

  ## Test plan

  - [x] `bun test src/` — 2171/2171 pass
  - [x] `bun run build` — clean
  - [x] `npx tsc --noEmit` — clean
  - [x] **RED-GREEN verified**: with the sticky branch disabled, 3 of 5 new tests fail (`always returns the same account for a given
  threadId`, `does not advance roundRobinIndex when sticky path serves the request`, `falls back to round-robin when sticky pick is
  cooling, then restores once cooldown lifts`). The other 2 lock pre-existing contract behavior (`preferredId still wins over threadId
  binding`, `returns null when only account is cooling, even with threadId`).

### DIFF
--- a/src/claude/account-pool.test.ts
+++ b/src/claude/account-pool.test.ts
@@ -1,7 +1,7 @@
 /**
  * Tests for AccountPool.
  */
-import { describe, it, expect } from 'bun:test';
+import { describe, it, expect, setSystemTime } from 'bun:test';
 import { AccountPool } from './account-pool.js';
 
 describe('AccountPool', () => {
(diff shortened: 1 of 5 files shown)


anneschuth:
Reviewed in detail — this is a solid structural fix and the tests are thoughtful (the n=2 vs n=3 reasoning and the `setSystemTime` choice over re-`markCooling` both stand out). Happy to merge.

Could you rebase against `main`? Just released v1.9.4 with the memory-leak fix in #356 and want to make sure this lands green against latest.

Two non-blocking observations for a possible follow-up:

1. **Hash distribution under monotonic threadIds.** With small `n` and threadIds derived from `<unix>.<counter>`-style timestamps (Slack thread_ts), `hash(threadId) % n` can bias toward whichever bucket the low-order bits favor. Worth a quick simulation in the next pass; not a blocker.
2. **Telemetry on sticky-vs-RR selection.** A `log.debug` in each branch would let us validate in production what fraction of acquires take the sticky path and what fraction fall through to RR. Helps confirm the fix is actually fixing rather than masking.

Neither needs to land in this PR.

shaders:
rebased

PR #352, opened 2026-04-28
feat: include thread permalink in Claude session context

shaders:
## Why

  The bot already gives Claude platform context (`**Platform:** … | **Working Directory:** …`) at the top of every session's system
  prompt, but never tells Claude the URL of the thread it is running inside. As a result, when Claude opens a merge request, files a
  ticket, or otherwise produces an artifact for someone to review later, it cannot link back to the conversation that led to the
  change. Reviewers have to find the thread by hand.

  ## What

  Extend `buildSessionContext()` to append `**Thread:** <permalink>`. Both platform clients already expose `getThreadLink()`, so this
  is a pure plumbing change with no new external dependencies.

  Resulting prompt segment:

  ```
  **Platform:** Mattermost (Team) | **Working Directory:** /repo | **Thread:** https://chat.example.com/_redirect/pl/<threadId>
  ```

  URL formats (unchanged, already in production):
  - Mattermost: `${url}/_redirect/pl/${threadId}`
  - Slack: `${teamUrl}/archives/${channelId}/p${threadId}` (root permalink — Slack threadIds are the root `thread_ts`)

  ## Files

  | File | Change |
  |---|---|
  | `src/commands/system-prompt-generator.ts` | `buildSessionContext` now requires `threadId` and accepts a platform with
  `getThreadLink` |
  | `src/operations/commands/handler.ts` | pass `session.threadId` on `!cd` respawn |
  | `src/operations/worktree/handler.ts` | build session context locally on `!worktree` restart (also restores Platform/WorkingDir,
  which had been silently dropped on this path pre-existing) |
  | `src/commands/system-prompt-generator.test.ts` | 1 new RED-GREEN test + 2 existing tests updated |
  | `src/operations/worktree/handler.test.ts` | mock platform exposes `platformType`/`displayName`/`getThreadLink` |

  ## Backward compatibility

  - No persisted-data shape changes.
  - No config changes.
  - `buildSessionContext` signature is breaking for any out-of-tree caller that passes only 2 args. All in-tree callers updated; tests
  cover the full call-site set.

  ## Test plan

  - [x] `bun test src/` — 2167/2167 pass
  - [x] `npx tsc --noEmit` — clean
  - [x] `bun run build` — clean
  - [x] **RED-GREEN verified**: with the `**Thread:**` segment removed from `buildSessionContext`, the new test (`includes the Thread
  permalink so Claude can reference the conversation`) fails on both the `getThreadLink` call assertion and the URL substring
  assertion.

### DIFF
--- a/src/commands/system-prompt-generator.test.ts
+++ b/src/commands/system-prompt-generator.test.ts
@@ -77,11 +77,20 @@ describe('generateChatPlatformPrompt', () => {
 });
 
 describe('buildSessionContext', () => {
+  const mattermostPlatform = {
+    platformType: 'mattermost',
+    displayName: 'Test Server',
+    getThreadLink: (threadId: string) => `https://chat.example.com/_redirect/pl/${threadId}`,
+  };
+
+  const slackPlatform = {
+    platformType: 'slack',
+    displayName: 'Workspace',
+    getThreadLink: (threadId: string) => `https://slack.example.com/archives/C123/p${threadId}`,
+  };
+
   it('formats platform and working directory', () => {
-    const context = buildSessionContext(
-      { platformType: 'mattermost', displayName: 'Test Server' },
-      '/home/user/project'
-    );
+    const context = buildSessionContext(mattermostPlatform, '/home/user/project', 'thread-1');
 
     expect(context).toContain('Platform:');
     expect(context).toContain('Mattermost');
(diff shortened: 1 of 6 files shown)


anneschuth:
Reviewed in detail — clean, well-scoped, RED-GREEN test is genuine. Happy to merge.

Could you rebase against `main`? Just released v1.9.4 with the memory-leak fix in #356.

Two non-blocking notes:

1. **`appendSystemPrompt` undefined-stringification risk in worktree handler.** The new code does ``\`${sessionContext}\n\n${options.appendSystemPrompt}\``` — if `options.appendSystemPrompt` is ever `undefined` (the type allows it), the literal `"undefined"` ends up in the prompt sent to Claude. In production it's always set from `CHAT_PLATFORM_PROMPT` so this is theoretical, but a `?? ''` guard would close the door cheaply.
2. **Bundled pre-existing-bug fix.** The "restore platform/working-dir context on worktree restart" fix is the right thing to do, but it's a separate concern from the thread permalink work. Splitting it into two commits (or two PRs) would be more bisectable. Not blocking — fine to keep folded in.

The reworked "capitalizes platform type" assertion is correct (the URL contains lowercase `slack`), though splitting on `|` is brittle to future segment reordering. A `expect(context).toMatch(/Platform:\\*\\* Slack \\(/)` would be sturdier. Minor.

shaders:
rebased

PR #366, opened 2026-05-05
Add read_post MCP tool to follow chat permalinks

anneschuth:
## Summary

`read_post` MCP tool resolves a Mattermost or Slack permalink to its content (and optional thread context), so Claude can act on links shared in the thread instead of asking you to copy-paste the message body. Auto-approved like `send_file`: the URL host + channel guards inside the handler are the real gate, since the bot's token is already scoped to what it can see.

Defaults: 20 thread messages per call (hard cap 50), 2000-char body truncation per message. Returned content is annotated as untrusted user input (prompt-injection note in the tool description).

## ⚠️ Breaking change

The MCP package name and server filename were renamed in this PR:

- MCP package id: `claude-threads-permissions` → `claude-threads-mcp` (matches the npm `bin` name and the broader scope: permissions + send_file + read_post)
- Server filename: `src/mcp/permission-server.ts` → `src/mcp/mcp-server.ts`
- Tool names exposed to Claude: `mcp__claude-threads-permissions__*` → `mcp__claude-threads-mcp__*`

In-flight sessions that span an upgrade will see new tool names on restart. The bot itself rewires automatically; no persisted data format changed. If you have external automation that grepped the old names (auto-approve allowlists, log filters), update accordingly.

## Commits

1. **Rename `PermissionApi` → `McpPlatformApi`** (`559fe78`) — pure rename, no behavior change.
2. **Mattermost `read_post`** (`5563a62`) — `readPost` / `readThread` on the API, Mattermost impl, `src/platform/mattermost/permalink.ts`, MCP tool registration.
3. **Slack `read_post`** (`a8d5afb`) — Slack implementation of the same shape. Resolution scoped to the bot's configured channel.
4. **First-pass review fixes** (`77c7a79`) — shared utilities, parser tightening, security framing, channel-scoping (initial pass).
5. **Second-pass review fixes + MCP rename** (`ccb698c`) — see below.

## Second-pass review fixes (`ccb698c`)

The first-pass fix introduced a regression: putting the Mattermost channel-scope check in the API layer meant the resolver couldn't distinguish "wrong channel" from "not found." Surface area was wrong. This commit moves the check up:

- `McpPost` gains a `channelId` field; resolver does the comparison itself; new `wrong-channel` `ResolveError` kind; handler renders it as the same "permalink is for a different channel" string Slack has been using.
- Mattermost handler gains the empty-`channelId` early-out Slack already had.
- `McpPlatformApi.readPost` simplified — no more `expectedChannelId` option (Slack ignored it anyway).

Other fixes from the second review:

- `TEAM_NAME_RE` now requires the first AND last character to be alphanumeric (mirrors Mattermost's server-side rule). Negative tests for leading/trailing hyphens and underscores; positive test for single-char teams.
- `expandTimestamp` inlined into `parseSlackPermalink` — one caller, no reason for the helper. `PATH_TS_RE` already enforces 16+ digits.
- `fetch-harness` now documents the array-identity contract explicitly: the `calls` array is mutated in place; future code that re-assigns it would silently break destructuring callers.
- MCP rename done end-to-end (filename + package id + tool names + build script + CI smoke-test paths + docs + system prompt + contract test reference).
- `CLAUDE.md` and `CHANGELOG.md` updated to match the new state.

## Out of scope (follow-ups)

- **Integration test for `read_post`.** The existing integration suite uses a mock Claude CLI; adding a `read_post` case needs new fixtures.
- **Slack workspace verification.** Currently the parser checks `*.slack.com` host but doesn't verify the workspace subdomain matches the bot's team. Channel-id collision across workspaces is the residual risk; channel-id check catches it in practice.

## Test plan

- [x] Parser shape tests on both platforms (Mattermost: `/pl/`, `/_redirect/pl/`, team-name validation including leading/trailing-hyphen/underscore rejection, query strings, wrong host; Slack: `/archives/{C}/p{ts}`, `?thread_ts=`, non-`*.slack.com` hosts, 16+-digit timestamp requirement).
- [x] Resolver tests: thread cap, ordering, channel scoping (Mattermost wrong-channel as a distinct error + Slack botChannelId), not-found / wrong-channel / unsupported error paths.
- [x] HTTP-level tests for the new MCP API methods (Mattermost `/posts/{id}` + `/posts/{id}/thread` with username caching; Slack `conversations.history` + `conversations.replies`).
- [x] MCP handler tests for both platforms including auto-approval, wrong-host / wrong-channel / not-supported error strings, the empty-channelId early-out, and the wrong-channel-vs-not-found distinction.
- [x] Parser logic RED-green-verified by mutation (3 targeted mutations per platform, first round).
- [x] Full unit suite: 2422 pass / 0 fail. ESLint clean. `tsc --noEmit` clean. `bun run build` green.
- [ ] Manual smoke test in #annes-claude-code-sessies after deploy: (1) plain permalink, (2) `include_thread`, (3) permalink for a different channel (must say "different channel," not "not found"), (4) deleted post.


anneschuth:
Review-feedback adressed in 77c7a79. Summary of what changed:

**Showstoppers**
- `parseMattermostPermalink` no longer accepts `/foo/pl/{id}` for arbitrary `foo` — segments[0] must be either `_redirect` literally or a valid Mattermost team-name (lowercase a-z, 0-9, hyphen, underscore; ≤64 chars). New negative tests for `/foo/bar/pl/{id}`, `/INVALID/pl/{id}`, `//pl/{id}`.
- Mattermost `read_post` now scopes to the bot's channel via `expectedChannelId` on `McpPlatformApi.readPost`. A permalink for a post in another channel the bot can technically see now returns null, matching the Slack behavior.

**Security**
- Tool description gains an explicit prompt-injection warning: fetched content is untrusted input, treat as data not instructions.
- Auto-approve branch documents why there's no `isUserAllowed` check at the MCP layer (the session-allowlist upstream is the real gate) and where the assumption breaks if read_post is ever invoked outside a session.

**API hygiene**
- `McpPost.createAt` documented: ms-since-epoch, best-effort across platforms (Slack drops microsecond precision); safe for sorting within one platform, not round-trippable. Re-fetch via `id`.
- Slack `PATH_TS_RE` tightened from 10+ digits to 16+ digits, folding the validation that was previously split between the regex and `expandTimestamp`.

**Refactor (no behavior change)**
- `src/platform/permalink-shared.ts` holds the cross-platform constants and helpers (`DEFAULT_THREAD_LIMIT`, `MAX_THREAD_LIMIT`, `MAX_MESSAGE_BODY_CHARS`, `clampThreadLimit`, `truncateBody`, `quoteBlock`).
- `src/platform/test-helpers/fetch-harness.ts` consolidates the fetch-recorder pattern that lived in three test files.

**Open**
- Live smoke test in #annes-claude-code-sessies (review item 3) — running below.
- "Permission server → MCP server" rename (review item 4, the half-renamed-name) — held for a separate PR; not the same scope as this one.
- An integration test for `read_post` (review item 10) — held as follow-up; the existing integration suite uses a mock CLI, adding a `read_post` case would require new fixtures.

All 2418 unit tests pass, lint + tsc clean.

PR #405, opened 2026-06-04
Add "respond only when mentioned" quiet mode (!mentions)

### LINKED ISSUE #402: Add "Respond only when mentioned" toggle
**Problem**
In threads that were started with claude-threads, users must explicitly mention each other in every reply to avoid claude-threads responding. If someone forgets to mention the other person, claude-threads treats the message as directed at it and replies. This makes it disruptive to hold side conversations within these threads.
**Proposed solution**
Add a per-thread configuration toggle: "Respond only when mentioned." When enabled, claude-threads will reply only to messages that explicitly @mention it. The current default behavior remains unchanged, so existing threads are unaffected — this is strictly opt-in for users who want quieter participation from claude-threads.
**Alternatives considered**
The only workaround today is to @mention the other participant on every message, since claude-threads ignores messages addressed to another user. This is tedious in active conversations and easy to forget, which is what motivates the toggle.

anneschuth:
Closes #402.

## Problem

In a thread started by claude-threads, every reply from an approved user is treated as input to Claude. To hold a side conversation with another participant, you currently have to @mention that person on every message (the bot ignores messages addressed to someone else). Forget once and the bot jumps in. This makes side conversations in an active thread tedious.

## Solution

A per-session, opt-in toggle: `!mentions`.

- `!mentions on` — the bot only responds to replies that explicitly @mention it. Other replies are ignored the same way side conversations to other users already are.
- `!mentions off` — back to default (respond to every approved-user message).
- `!mentions` (no argument) — flips the current value.

Commands are evaluated before the gate, so `!mentions off` always works even while quiet mode is on. The default is unchanged, so existing threads behave exactly as before. Authorization matches other per-session settings (session owner or a globally allowed user), and the flag persists across a bot restart.

## How it works

The gate lives in `handleMessage` (`src/message-handler.ts`), right after command parsing and before the message is sent to Claude: if `session.respondOnlyWhenMentioned` is set and the incoming reply does not `isBotMentioned(...)`, the handler returns early. When the bot is mentioned, the existing `extractPrompt` path strips the mention so Claude receives a clean prompt.

State is a single boolean `respondOnlyWhenMentioned` threaded through the standard places: `Session` (in-memory), `PersistedSession` (disk, optional with a `?? false` default on resume for backward compatibility), defaulted to `false` on session creation, written in `manager.ts` persist, restored on resume.

## Files

- `src/session/types.ts`, `src/persistence/session-store.ts` — the flag on `Session` and `PersistedSession`.
- `src/session/lifecycle.ts` — default `false` on create, `?? false` on resume.
- `src/session/manager.ts` — persist the flag, delegate `setRespondOnlyWhenMentioned`.
- `src/operations/commands/handler.ts` — the toggle handler (auth, mutate, persist, confirm, update header).
- `src/commands/registry.ts`, `src/commands/parser.ts`, `src/commands/executor.ts` — register/parse/dispatch `!mentions`. Help text is auto-generated from the registry.
- `src/message-handler.ts` — the routing gate.

## Tests

- `message-handler.test.ts` — gate behavior: ignored without mention, responds with mention, default-off path, and commands (including `!mentions off`) bypassing the gate. Verified RED: removing the gate fails exactly the gate-dependent test.
- `operations/commands/handler.test.ts` — on/off/toggle and the unauthorized-user rejection (no state change, no persist).
- `commands/parser.test.ts` — parsing of `on`, `off`, and the bare toggle.
- Backward-compat: the persisted-key-set assertion in `manager.test.ts` updated; existing `Session` fixtures gained the new required field.

Full suite, typecheck, and lint pass. (One unrelated pre-existing failure, `chmod-portable › lands the SGID bit`, fails on a clean tree too; it is environment-specific and untouched by this change.)


roskee:
@anneschuth thanks for working on this. I just wanted to mention that there should be a global kind of config to change the default value of this toggle for all future threads. otherwise one would have to run `!mention on` every time they want to hold a side conversation on a thread.

anneschuth:
Good call, thanks. Added a global `respondOnlyWhenMentioned` option in `config.yaml` (`d1737e3`) so you can default new threads to quiet mode without running `!mentions on` each time. The onboarding wizard prompts for it too (first-time setup and reconfigure), and it shows in the config summary. Each session still keeps its own value, so `!mentions` can override per-thread, and resumed sessions retain whatever they had. Default stays `false`, so existing configs and threads are unaffected. Docs updated in CONFIGURATION.md, the README command table, and CLAUDE.md.

anneschuth:
Did another thorough review round (correctness/edge-cases and persistence/backward-compat). Two things worth fixing surfaced, both addressed in 47c756b:

1. **Worktree prompt + quiet-mode default could silently drop the branch name.** With the new global `respondOnlyWhenMentioned: true`, a fresh session starts in quiet mode *and* gets a worktree prompt. Answering it is a plain reply with no @mention, which the gate dropped, so the prompt looked ignored. The pending-worktree-prompt handler now runs **before** the quiet-mode gate, the same way commands already bypass it. Regression test added and RED-verified (fails with the old ordering).

2. **Quiet mode wasn't discoverable.** It was only visible in the one-time confirmation message. The session header now shows a `🔕 Replies: only when @mentioned (!mentions off to disable)` row while quiet mode is on, so a returning user can see why the bot is staying quiet. (This also makes the `updateSessionHeader` call in the toggle handler meaningful instead of a no-op re-render.)

Also added a backward-compat test that an old `sessions.json` lacking the field loads cleanly (resume defaults it to false).

The rest checked out: resume correctly uses the session's own persisted value and ignores the config default; commands and side-conversation ordering are right; the persisted-keys drift assertion covers the new field; old configs/sessions load without it. Full suite, typecheck, and lint pass (the one unrelated pre-existing `chmod-portable` SGID failure fails on a clean tree too).

PR #437, opened 2026-07-24
Add opt-in per-message user attribution (userAttribution flag)

bormog:
## What

Add an **opt-in** `userAttribution` config flag (default `false`). When enabled,
every genuine user turn sent to Claude is prefixed with the sender's
`[@username]:` (their platform login) right before the message is handed to
Claude, so Claude can tell who is speaking in a thread shared by several people.

With the flag off — the default — behaviour is byte-for-byte unchanged:
upgrading produces exactly the same prompts and system prompt as before.

## Why

In a multi-participant thread (`!invite`, message approval), every turn reached
Claude anonymously, so Claude could not attribute requests to the person who
made them. Prefixing each turn with the sender fixes that without changing the
single-user experience.

## How

- **Composed only at the send boundary.** A small pure `formatUserTurn(message,
  username, enabled)` in `src/operations/user-attribution/` wraps the message
  immediately before it goes to Claude. The sender identity is carried
  separately and is never baked into the stored prompt, so the prefix never
  leaks into thread titles, git branch-name suggestions, or persisted session
  state. This mirrors the existing `formatApprovedMessage` /
  `formatSideConversationsForClaude` idiom.
- **Every real send path is covered:** in-thread follow-ups, resumed turns, the
  initial mid-thread prompt, post-`!cd` re-sends, the thread-context-prompt
  paths, and worktree re-sends. A paused session's reviving message is
  attributed to its actual sender, not the owner. Side-conversation context is
  kept outside the `[@username]:` prefix so other users' remarks aren't credited
  to the current speaker.
- **System / control sends are deliberately left unattributed:** slash-command
  passthrough, plan approval, and question/approval completion.
- **One system-prompt line, gated on the flag.** When attribution is on, a
  system-prompt note tells Claude to treat the prefix as speaker metadata and
  not to echo it in replies or commit messages. When off, the note is omitted
  too, so Claude is never taught a prefix that never arrives. The append is
  rebuilt when `!permissions` respawns Claude.
- **Backward-compatible persistence.** The flag is persisted per session, so a
  session keeps its behaviour across bot restarts. Sessions persisted before
  this change read as off. Applies to NEW sessions; a resumed session keeps the
  value it started with.
- **Onboarding** offers the opt-in as a question; `config.yaml`, `CLAUDE.md`,
  and `docs/CONFIGURATION.md` document it.

## Testing

- Unit tests per send path with RED-GREEN discipline (formatter, system-prompt
  generator, message-manager, worktree, context-prompt, lifecycle, onboarding).
- Multi-user integration test: a second user's follow-up reaches Claude
  attributed; integration coverage runs behind the flag and asserts the default
  stays raw.
- Full suite green: `bun test` (2720 pass / 0 fail), `bun run lint`,
  `bun run typecheck`.
- Manually e2e-tested on a live instance with `userAttribution: true`.

### Note on the `knip` CI check

A recent `knip` release (knip is not pinned — CI runs `bunx knip`, which pulls
the latest) started flagging pre-existing barrel re-exports across the repo, so
the `knip` check is currently red on `main` as well (86 unused exports on
`upstream/main` at `cb1fb50`). This branch is **net-zero** against that baseline
— it introduces no new unused exports (the one it did add, an internal
`sanitizeUsername` re-export, has been removed). Fixing the knip drift is a
separate, repo-wide infra change and is intentionally out of scope for this
feature PR.

## Known limitation (pre-existing, out of scope)

When a session starts mid-thread with prior context, the bot posts a
"include previous conversation?" prompt and defers the queued message,
persisting it. If the bot restarts while that prompt is still pending, the
deferred state is not rehydrated on resume (`resumeSession` never forwards
`pendingContextPrompt`), so the eventual reaction is dropped and the queued
message — including its attributed sender — is lost. This is a pre-existing
persistence bug independent of attribution; it is intentionally **not** fixed
here (the proper fix needs context-prompt timeout re-arming and belongs on its
own). The CHANGELOG and code comments avoid claiming deferred-context restart
survival.


anneschuth:
Thanks for your contribution, this is a good idea. And  possibly the default should actually be set to true. 

I'll look into merging and fixing CI (knip) when I'm back from vacation.

anneschuth:
Thanks for this, @bormog — the feature and nearly all of the implementation are yours.

It's shipped in **v1.19.0** via #446, which carries your commits forward with authorship intact and credits you as co-author on the squashed merge. #446 rebased onto current `main`, flipped the default on (gated so only multi-participant threads get the prefix), and added a docs-consistency pass — the design and code are yours.

Closing this as superseded, not rejected. Thanks again for the contribution.

---

PR #446, opened 2026-07-27
Per-message user attribution, defaulted on for shared threads (supersedes #437)

anneschuth:
Carries @bormog's #437 forward. Their 24 commits are preserved with authorship intact — this branch exists only because #437's head lives on a fork that this session cannot push to, so #437 itself could not be updated in place. **#437 should be closed as superseded, not as rejected** — the feature and nearly all of the code are theirs.

## 1. Rebased onto current main

#437 was based on `cb1fb50`, before the CI repairs, and had gone `dirty`. Merged onto current main (v1.18.4); only `CHANGELOG.md` conflicted, and both sections are kept.

**The knip check that #437 called out as red is now green.** Its description scoped that fix out because main was red too; main now pins knip and treats `src/**/index.ts` as entry points, so the new user-attribution barrel isn't flagged. Nothing about the feature had to change for it.

## 2. Default on, but only for shared threads

Per review feedback on #437, the default flips `false` → `true` — but a flat default-on would put `[@username]:` in front of every message in solo threads, where the prefix names the only person who could have spoken. So the prefix is gated on the session actually having more than one participant.

The signal is the existing session allowlist: seeded `new Set([owner])` and grown only by real collaboration events (`!invite`, or another user reviving a paused session). No new state. It's evaluated **per send**, not once at session start, so a thread that becomes shared mid-session starts attributing from that point on.

```ts
export function shouldAttribute(enabled: boolean, participantCount: number): boolean {
  return enabled && participantCount > 1;
}
```

Kept as a pure predicate over primitives so it stays trivially testable and pulls no `Session` import into the attribution module. All six send sites route through it; `formatUserTurn` itself is untouched.

Two details worth reviewing:

- **The system-prompt note keys off the config flag, not the live participant count** — so Claude is primed for a prefix that may only start arriving once someone is invited mid-session.
- **`pruneDefaultFalseFlags` no longer drops `userAttribution`.** With a `true` default, pruning an explicit `false` would silently re-enable the feature on the next start, making opt-out impossible. This is the one place the flipped default has a non-obvious consequence.

Resumed sessions keep the value they started with; sessions persisted before the flag existed stay unattributed.

## Verification

- 4 new tests for the gate, **verified RED before GREEN** — removing the participant check fails 2 of them, restoring it passes all 14.
- 8 existing tests updated for the new semantics: three asserted the old default, five enabled the flag on solo sessions the gate now correctly suppresses.
- Full suite green: **2650 pass / 0 fail**, plus typecheck, lint, knip, build and `bun audit`.

## Not addressed

The pre-existing deferred-context persistence limitation documented in #437 is still out of scope and unchanged here.

https://claude.ai/code/session_01J3XKJQj5RJwiBZvyLYfc7M

---
_Generated by [Claude Code](https://claude.ai/code/session_01J3XKJQj5RJwiBZvyLYfc7M)_


anneschuth:
Merged and shipped in **[v1.19.0](https://github.com/anneschuth/claude-threads/releases/tag/v1.19.0)** — `release.yml` tagged, released, and published to npm. @bormog is preserved as co-author on the squashed merge; the feature and the bulk of the code are theirs (#437, now closed as superseded).

Landed on top before release: a docs/comment consistency pass (the CHANGELOG, `CONFIGURATION.md`, `CLAUDE.md`, and several code comments still described the old default-off behavior — now all say default-on, gated to shared threads), plus the `1.19.0` minor bump.

---

PR #467, opened 2026-08-08
Persist TaskTracker state across bot restarts (1.24.1)

anneschuth:
## What

Fixes task-name loss after a bot restart. Modern CLIs stream tasks incrementally (`TaskCreate`/`TaskUpdate` with ids), and the bot accumulates them in a per-session `TaskTracker` — but that tracker lived only in memory. After a restart + resume, the first `TaskUpdate` of the next turn hit an empty tracker and rendered a **"Task #1" placeholder** instead of the real subject, degrading the task list for the rest of the session.

## How

- **`TaskTracker.serialize()` / `restore()`** (`src/operations/task-tracker.ts`): resolved tasks round-trip as `{taskId, subject, activeForm?, status, isPlaceholder?}`. In-flight creates (id not yet resolved from the `"Task #N created successfully"` result text) are deliberately dropped at serialize time. `restore()` validates defensively: entries without an id/subject are skipped, unknown statuses fall back to `pending`.
- **`PersistedSession.taskTrackerState?`** (`src/persistence/session-store.ts`): new optional field. Absent on pre-1.24.1 data → tracker starts empty, exactly today's behavior (backward-compat rules followed: additive field, defensive read).
- **Turn-end persistence** (`src/operations/events/handler.ts`): the `result` event branch now calls `persistSession`, so tracker state reaches disk at every turn end instead of waiting for the next lifecycle write. The CLI only runs while the bot runs, so a turn-end snapshot cannot go stale.
- **Resume wiring** (`src/session/lifecycle.ts`): the tracker is restored right after `restoreTaskListFromPersistence`.

## Testing (red-green verified)

- **Unit** (7 new tracker tests + wiring tests in `manager.test.ts` / `events/handler.test.ts`): written first and confirmed failing without the implementation. The persistence snapshot guard test now covers `taskTrackerState`.
- **Integration** (`session-resume.test.ts` + new `task-list-partial` mock scenario): starts a session that creates two tasks and leaves them incomplete, restarts the bot preserving `sessions.json`, drives the next turn (mock resumes its scenario state and emits `TaskUpdate` → completed), and asserts the re-rendered list shows `✅ ~Analyze requirements~` with no `Task #1` placeholder. **Verified RED** with the lifecycle restore call removed (fails in 35s), GREEN with it (passes in 8s). The scenario's `result` steps carry a 5s delay because `finalize` deletes an incomplete task-list post at turn end — the test needs a window to observe the render.
- Full unit suite: 2887 pass. Both task/resume integration suites pass on the Slack path locally; CI runs both platform matrices.

## Release

Carries the version bump to **1.24.1** + CHANGELOG entry — merging to main releases it (patch: pure fix, no API change).

---
_Generated by [Claude Code](https://claude.ai/code/session_01KhTYe3y4UgG2Bx8kGRyo4K)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.24.1] - 2026-08-08
+
+### Fixed
+- **Task lists no longer degrade to "Task #N" placeholders after a bot restart.** Modern CLIs stream tasks incrementally (`TaskCreate`/`TaskUpdate` with ids), and the bot accumulates them in a per-session `TaskTracker` — but that tracker lived only in memory. After a restart + resume, the first `TaskUpdate` of the next turn hit an empty tracker and rendered a placeholder ("Task #1") instead of the real subject, losing every task name for the rest of the session. The tracker's resolved tasks (id → subject/status) are now persisted to `sessions.json` at each turn end and restored on resume. In-flight creates (id not yet resolved from result text) are deliberately dropped at serialize time; pre-1.24.1 persisted sessions simply start with an empty tracker as before. Covered by a red-green integration test that restarts the bot mid-task-list and asserts the post-resume re-render still shows real subjects on both platform paths.
+
 ## [1.24.0] - 2026-08-08
 
 ### Added
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "claude-threads",
-  "version": "1.24.0",
+  "version": "1.24.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "claude-threads",
-      "version": "1.24.0",
+      "version": "1.24.1",
       "license": "Apache-2.0",
       "dependencies": {
         "@hono/node-server": "2.0.12",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "claude-threads",
-  "version": "1.24.0",
+  "version": "1.24.1",
   "description": "Run Claude Code from Slack or Mattermost. Sessions stream live into threads where your whole team can watch and steer.",
   "main": "dist/index.js",
   "type": "module",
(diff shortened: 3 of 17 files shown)


anneschuth:
**Adversarial review round 1 — done, all findings fixed in 115d652.**

| # | Severity | Finding | Fix |
|---|----------|---------|-----|
| 1 | MEDIUM | Turn-end persist ran **before** the result event's op chain finished — `taskListExecutor.finalize()` (which deletes an incomplete task post and nulls its state) runs inside that chain, so every such turn persisted a `tasksPostId` pointing at a just-deleted post. | Post-processing now receives the `MessageManager.handleEvent` promise and chains the persist after it settles (post-finalize snapshot). Red-green unit tested with a controllable promise. |
| 2 | MEDIUM | `persistSession` could throw (ENOSPC/EACCES on the sync `sessions.json` write) inside the CLI `'event'` listener chain, where the JSON-parse catch swallowed it as stream noise **and aborted the rate-limit scan for that same event** — error results are exactly the events carrying rate-limit signals. | `persistSession` logs instead of throwing; `parseOutput` now separates parse errors (ignored) from listener errors (logged), so the rate-limit scans always run. Red-green tested with a throwing listener. |
| 3 | LOW | `TaskTracker.restore` threw on null/primitive rows or a non-array `taskTrackerState` — and a restore throw aborts the entire session resume. | Fully defensive now: non-array resets to empty, non-object rows dropped. Red-green tested with `[null, 'junk', 42, valid]` and a string value. |
| 4 | LOW | The integration test didn't exercise the turn-end persist (the graceful-shutdown persist masked its removal), and the resume-wait loop fell through without an assertion. | Test now reads `sessions.json` after turn 1's result — before any shutdown persist — and asserts the tracker snapshot is on disk (red-verified against a disabled turn-end persist); explicit assertion when the session fails to come back; scenario delays 5s → 8s for CI-load margin. |
| 5 | LOW | Per-turn persist is a synchronous full read-modify-write of `sessions.json`. | Acceptable (atomic temp+rename, single-process); documented at the call site. |

Verified-clean angles from the review (no action needed): create-resolution ordering vs the turn's result, `pendingCreates` clearing, `'deleted'` status round-trip, placeholder/allCompleted guard, `!cd` respawn clearing, forward/backward compat, `STORE_VERSION` untouched, TodoWrite interplay.

Also in this PR since the last CI run: the Mattermost matrix failure was a harness gap — the test harness mints a fresh platform id per bot start, so a restarted bot could never resume persisted sessions (keyed by `platformId:threadId`). Production ids come from config.yaml and are stable; `platformIdOverride` now carries the id across the test restart (d7945d5).

---

PR #477, opened 2026-08-19
fix: show the full Bash command in permission prompts

Jadefalkner:
### Problem

The permission prompt hard-truncates Bash commands at 100 characters (display line: 50). Anything past the first pipe or `&&` is invisible at the exact moment the user is asked to approve it:

> ⚠️ **Permission requested**
> 💻 **Bash** `npm view claude-threads version 2>/dev/null; echo "---issues---"; gh search issues --repo anneschuth...`
> 👍 Allow | ✅ Allow all | 👎 Deny

The user approves the *beginning* of a command. The part that matters — what comes after the `...` — is exactly the part they cannot see, so the gate can only be rubber-stamped.

### Change

- The permission prompt renders the command as a fenced ```bash code block, up to a generous 1500-character cap (`PERMISSION_COMMAND_MAX`). A pathological command is still cut (marked `[... truncated]`) so it cannot blow up the prompt post; the platform layer's overall message-size limits still apply on top.
- Fenced rendering also closes an injection surface: a command containing backticks could previously break out of the inline code span and corrupt the prompt, including the reaction legend. `escapeCodeBlockContent` is now convergent (a single pass turned four backticks into a fresh ``` fence; it now repeats until no triple-backtick run remains).
- Truncation cuts at Unicode code-point boundaries, so it can no longer split a surrogate pair.
- The 50-character display truncation in the streaming view is unchanged — only the permission prompt is affected.

### Tests

- full command shown / no ellipsis below the cap
- pathological 5000-char command still capped
- backtick runs (including 5+ backticks) cannot recreate a fence inside the block
- astral-emoji command truncates without producing replacement characters

All verified red-green (each test fails with the fix reverted).

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+- **Permission prompts show the full Bash command.** The approval prompt used to hard-truncate commands at 100 characters, so anything past the first pipe or `&&` was invisible at the exact moment the user was asked to approve it — the gate could only be rubber-stamped. The prompt now shows the command up to a generous 1500-character cap (a pathological command is still cut so it cannot blow up the prompt post). The 50-character display truncation in the streaming view is unchanged; only the permission prompt is affected.
+
 ## [1.26.0] - 2026-08-20
 
 ### Added
--- a/src/operations/tool-formatters/bash-tools.ts
+++ b/src/operations/tool-formatters/bash-tools.ts
@@ -15,6 +15,10 @@ import { escapeRegExp, escapeCodeBlockContent } from './utils.js';
  * the command up to this cap; anything longer is still cut so a pathological
  * command cannot blow up the prompt post (the platform layer enforces its own
  * overall message-size limits on top).
+ *
+ * The cap counts Unicode code points, not UTF-16 units: an all-astral command
+ * can emit up to twice this many units (~3000), still far under the platform
+ * layer's 16K message split.
  */
 const PERMISSION_COMMAND_MAX = 1500;
 
--- a/src/operations/tool-formatters/bash-tools.ts
+++ b/src/operations/tool-formatters/bash-tools.ts
@@ -9,6 +9,15 @@
 import type { ToolFormatter, ToolFormatResult, ToolInput, ToolFormatOptions } from './types.js';
 import { escapeRegExp } from './utils.js';
 
+/**
+ * The permission prompt is the user's only chance to see what is about to run —
+ * a hard-truncated command turns the approval gate into rubber-stamping. Show
+ * the command up to this cap; anything longer is still cut so a pathological
+ * command cannot blow up the prompt post (the platform layer enforces its own
+ * overall message-size limits on top).
+ */
+const PERMISSION_COMMAND_MAX = 1500;
+
 // ---------------------------------------------------------------------------
 // Bash Formatter
 // ---------------------------------------------------------------------------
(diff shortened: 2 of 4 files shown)


anneschuth (changes requested):
Thank you for this — it fixes a real weakness at exactly the spot where visibility matters most, and the write-up made it a pleasure to review.

I verified the core claims independently and they hold up well:

- **The fence-breakout fix is sound.** The old single-pass `escapeCodeBlockContent` was genuinely exploitable (four backticks → `` ` ``` `` recreates a fence, letting a crafted command close the block and append fake markdown to the permission post). I tested the convergent loop empirically for backtick runs of length 3–30: no triple-backtick run survives, and it converges in ≤3 passes, so there's no pathological-input concern.
- Truncation marker inside the fence, after escaping — can't be used to break out. 👍
- `bun test src/operations` (1179 tests), `tsc --noEmit`, and eslint are all clean on the branch, including the other `escapeCodeBlockContent` callers in the file-tools previews.
- The convergence test is properly red-green: reverting `utils.ts` to single-pass makes it fail.

**One blocking ask** — the surrogate-pair test in `tool-formatters.test.ts` ("never cuts inside a surrogate pair") isn't actually red without the fix: replacing `truncateAtCodePoint` with a naive `cmd.substring(0, 1500)` still passes all 118 tests. Two independent reasons:

1. 1500 is even and the fixture is uniformly 2-unit emoji, so a naive cut lands on a pair boundary anyway.
2. A broken cut produces a *lone surrogate* (`\uD83D`), never `�` — the replacement character only appears on UTF-8 re-encode, so the `includes('�')` assertion can never fire in-process.

The production fix itself is correct and worth keeping — only the test is inert. Suggested rewrite: prefix the fixture with a single BMP char (`'a' + emoji.repeat(...)`) to force an odd boundary, and assert `result!.permissionText!.isWellFormed()` (available on Node 20+/Bun, within the repo's engine floor). That makes it genuinely fail when the fix is removed.

**Non-blocking nit:** `PERMISSION_COMMAND_MAX = 1500` is enforced in code points, so an all-astral command can emit up to 3000 UTF-16 units — harmless under the platform layer's 16K split, but the comment reads like a character cap; a one-line clarification would help the next reader.

With the test rewrite this is good to go. Thanks again!

---

Jadefalkner:
Thanks for the sharp catch — you're right on both counts: the even cap landing on pair boundaries *and* the lone-surrogate-vs-U+FFFD point made the test structurally unable to fail.

Rewrote it as suggested: the fixture is now `'a' + emoji.repeat(1600)`, forcing an odd-index cut for a naive substring. One deviation from your suggestion: the assertion uses `/[\uD800-\uDFFF]/u` (with `/u`, paired surrogates form a single code point outside that range, so the class matches only *lone* surrogates — equivalent to `!isWellFormed()`). `isWellFormed()` itself trips `tsc --noEmit` because the repo's tsconfig targets ES2022; bumping the target felt like scope creep for this PR, but happy to do that instead if you prefer.

Verified red-green by temporarily swapping `truncateAtCodePoint` for a naive `substring(0, 1500)`: 1 fail with the naive cut, 0 with the fix.

Also expanded the `PERMISSION_COMMAND_MAX` comment to state it counts code points (an all-astral command can emit ~3000 UTF-16 units, still far under the 16K platform split).

anneschuth:
Thanks for the quick turnaround! I verified the rewritten test is genuinely red-green (naive `substring` → exactly that test fails; with the fix → all 118 pass), and the `/u`-regex-instead-of-`isWellFormed()` reasoning is sound given the ES2022 target. The blocking item from the review is resolved.

One last mechanical step: the branch now conflicts with `main` — the 1.25.1 release just promoted the CHANGELOG's `## [Unreleased]` heading to `## [1.25.1] - 2026-08-20`, which collides with this PR's CHANGELOG entry. Could you merge `main` into the branch and re-seat your entry under a fresh `## [Unreleased]` section above `## [1.25.1]`? Once that's pushed and CI runs on the new head, this is ready to approve and merge.

---

Jadefalkner:
Rebased onto `main` (2ed7dd5) — the CHANGELOG entry is re-seated under a fresh `## [Unreleased]` above 1.26.0/1.25.1, no other changes. Conflicts are gone.

anneschuth (approved):
Approving — everything requested is done and independently verified on the rebased head (2ed7dd5):

- The surrogate-pair test is now genuinely red-green (re-verified after the rebase: naive `substring` → exactly that test fails; with the fix → 118/118 pass).
- The rebase is clean: vs `main` the PR contributes exactly the four expected files, and the CHANGELOG entry is correctly re-seated under a fresh `## [Unreleased]` above 1.26.0.
- `tsc --noEmit` clean on the branch.

Thank you for the careful fixes and the fast turnaround. Once CI runs on this head (needs a maintainer approval click), this is ready to merge.

---

Issue #315, opened 2026-04-08
Support DM channel mode: interact with Claude without @mention


meredith620:
### Summary

Add support for interacting with Claude Code in Mattermost Direct Message (DM) channels without requiring an @mention.

### Motivation

In Mattermost, DM channels are already a private 1-on-1 conversation between the user and the bot. Requiring an @mention in this context adds unnecessary friction — every message already implicitly targets the bot.

A more natural UX would allow users to simply type their request directly in the DM channel, without prefixing with `@bot-name`.

### Proposed Behavior

- When the configured channel is a Mattermost DM channel (channel type `D`), treat every incoming message as a bot mention
- No `@bot-name` prefix required to start a session or send follow-up messages
- All existing session commands (`!stop`, `!escape`, `!cd`, etc.) continue to work as-is
- Session persistence and resume behavior unchanged

### Implementation Notes

The `MattermostClient` already has an `isDirectMessage()` method that detects DM channels (type `D`) at connect time. The changes would be minimal:

1. `isBotMentioned()` — return `true` when `_isDmChannel` is `true`
2. `extractPrompt()` — return the full message (no @mention to strip) when `_isDmChannel` is `true`
3. In `message-handler.ts`, use `post.channelId` as `threadRoot` instead of `post.rootId || post.id` for DM channels, so all messages in the DM map to a single persistent session

We have already implemented and tested this on our fork and are happy to submit a PR if this feature is of interest to the project.

### Platform

- Mattermost (DM channel type `D`)
- Slack equivalent could be considered in a follow-up

anneschuth:
Thanks for the detailed proposal and implementation! DM channel support is an interesting feature, but it's a larger change that I'll consider for a future release. Closing for now — feel free to open a PR when you're ready and we can iterate on the implementation details there.

PR #478, opened 2026-08-19
feat: direct channel mode — the channel is the session (#315)

Jadefalkner:
### Motivation

Threads are the right isolation for parallel work, but for a dedicated conversation with the bot they add friction: every session lives behind a reply arrow, and on mobile Mattermost that is noticeably clunky. #315 asked for exactly this for DMs ("every message already implicitly targets the bot"). Direct channel mode (DCM) generalizes that idea: opt a channel in, and the whole channel becomes one continuous conversation.

I have been running this in production against a self-hosted Mattermost since day one of writing it — the examples below are from that setup.

### What it does

```yaml
platforms:
  - id: mattermost-main
    type: mattermost
    # ... regular entry ...
    directChannelMode: true   # shorthand; long form below
```

- No `@mention` needed — every message from an allowed user reaches the bot.
- The bot replies with **top-level channel posts** instead of thread replies; the channel reads like a plain chat.
- One session per platform entry (the channel *is* the session). Messages posted inside any thread of the channel route to that same session.
- `!commands`, emoji permission prompts, plan approvals, persistence and resume-after-restart all work unchanged.

### How it works

The session is keyed by a synthetic thread id (`dcm:<platform id>`) instead of a real post id. The platform clients resolve that id at the API boundary (`root_id` / `thread_ts` → channel root), so the entire session machinery — registry, persistence, reaction routing, the MCP permission server — is reused as-is. No parallel code path.

Places where a synthetic id must not leak are guarded individually: typing indicators, thread links, `!cd` work summaries, context-prompt history, and the no-URL MCP fallbacks (`list_thread` / `react_to_post` return a clear error and point to `read_channel_history` / permalinks).

### Options

```yaml
directChannelMode:
  respondTo: all_messages   # or: mention
```

`respondTo: mention` is for shared channels where several people discuss and the bot should only join when addressed. It seeds the existing per-session quiet-mode flag (#402), so `!mentions` still toggles it at runtime.

### Platform-level `approvals` (also useful without DCM)

```yaml
  - id: mattermost-main
    approvals: owner          # or: all_users
```

Controls who may answer tool-permission prompts and the other reaction gates (plan approvals, question answers, resume): `owner` = the session participants (starter + `!invite`d users), `all_users` = everyone on the entry's `allowedUsers`.

Unset keeps the historical default per mode — `all_users` for thread sessions (**unchanged upstream behavior**), `owner` for DCM (a shared channel should not let every allowed user approve by default). Setting `approvals: owner` on a classic thread channel is an opt-in hardening: today any platform-allowed user can approve tool use in anyone's thread session.

### DMs, the config-only way

On Mattermost a DM is just a private channel with a stable id, so a DCM entry pointed at it covers bot direct messages with zero extra code:

```yaml
  - id: mattermost-dm
    type: mattermost
    url: https://chat.example.com
    token: same-bot-token
    channelId: <dm-channel-id>   # POST /api/v4/channels/direct ["<bot-id>","<user-id>"]
    botName: claude-code
    directChannelMode: true
    stickyMessage: hidden
    allowedUsers: [you]
```

This is deliberately **Mattermost-only**: Slack Socket Mode distributes event envelopes across an app's active connections, so a second entry on the same credentials can consume and discard events meant for the first. (A follow-up PR adds dynamic DM discovery on top, so static entries are only needed if you want them.)

### Scope note: DCM and the approvals option are bundled

DCM needs an approvals story (a shared channel should not let every allowed user approve tool use), which is why the platform-level option ships in this PR. If you would rather take DCM alone, I am happy to split: either drop the configurable option and keep DCM's owner scoping fixed, or extract `approvals` into its own PR (it is useful for classic thread channels on its own). Say the word and I will restructure.

### Non-breaking

Everything is behind the opt-in flag; classic thread sessions are bit-identical to upstream — the existing ~2900 tests pass unchanged. `approvals` unset preserves today's behavior exactly.

### Limitations (documented)

- The thread-context prompt is skipped in DCM (no thread history behind a synthetic id).
- `list_thread` cannot resolve the synthetic session id (error message points to alternatives).
- The approval set is fixed at CLI spawn; a later `!invite` reaches it on the next respawn (e.g. `!cd`).

### Testing

Unit tests for the resolver, message routing (with/without mention, thread routing, unauthorized users, `respondTo: mention`), approvals scoping across fresh spawn / resume / respawn / reaction gates / resume-from-reaction, and DCM session resume. Hardened through fourteen adversarial review rounds (a second model hunting for leaks, races, and privilege bypasses); every finding is covered by a regression test. A real-Mattermost integration suite covers the DCM contract: no-mention session start, channel-root replies, thread routing into the channel session, `!stop`, and the `respondTo: mention` gate.

---

---

**Review follow-up note (isRunning() pre-check):** removing the `isRunning()` pre-check in `manager.ts` changes behavior for all platforms, not just DCM — deliberately so. The pre-check raced the asynchronous start: a session could be registered but not yet running, and the pre-check would treat it as dead and spawn a duplicate. The in-flight map now owns that window for every platform; `sendFollowUp` still verifies liveness (with a bounded wait when a start is in flight) before delivering.


anneschuth (changes requested):
First off: this is an impressively engineered PR, and the honest, thorough description made reviewing it far easier than a change of this size usually is. The synthetic-id containment is genuinely clean — I checked every platform egress I could find (`createPost` on both platforms, the typing indicator's `parent_id`, both MCP platform APIs, both file-upload paths, `getThreadLink` fallbacks) and found no leaks. `PersistedSession` is untouched, resume uses defensive defaults, the onboarding wizard round-trips the new options, and the full unit suite (3000 tests) plus typecheck pass on the branch. The tests drive real code paths rather than copied logic. Really nice work.

That said, the PR introduces a security option and currently ships two allowlist-level ways around it, so I have to request changes:

**1. Blocking: `!approve` bypasses `approvals: owner`.** The reaction path is correctly gated (`src/session/reaction-router.ts:106-112` via `resolveApprovals`), but the documented text alternative `!approve` (`handleApprove` in `src/commands/executor.ts`, → `approvePendingPlan` in `src/operations/commands/handler.ts`) only checks `ctx.isAllowed` — the platform allowlist. Under `approvals: owner`, any platform-allowlisted non-participant can approve a pending plan (including resolving the decision bridge) by typing `!approve`, which contradicts the docs' claim that plan approvals are owner-scoped. The fix looks small: apply `resolveApprovals` in `approvePendingPlan` (or in the command's gate) the same way the reaction router does.

**2. Blocking (or document the trust model): `!invite` self-escalation.** `requireSessionOwner` accepts any platform-allowlisted user, so a non-participant can `!invite` themselves, immediately passing the owner-scoped reaction gate (it reads live `sessionAllowedUsers`) and joining the MCP approval set on the next respawn. That reduces `owner` from a boundary to an audit trail. Either make `!invite`/`!kick` participant-only when the effective approvals mode is `owner`, or state explicitly in the docs that `owner` is auditable-but-not-enforced against allowlisted users.

Three non-blocking observations, fine as follow-ups:

- **Message-based resume isn't owner-scoped** while reaction-based resume is (`reaction-router.ts:159-170` vs. the paused-session path in `message-handler.ts`). Low severity — resuming isn't approving — but it undercuts the stated parity.
- **Memory distillation silently no-ops for DCM sessions**: `distillThread` calls `platform.getThreadHistory(threadId)` with the synthetic id (`src/memory/distiller.ts:126-128`) and the failure is swallowed by the fire-and-forget catch, so DCM sessions never feed channel memory. The other thread-history consumers were guarded; this one looks missed.
- **`_inFlightSessionStarts` has no tests** despite being exported "for tests only" (`src/session/lifecycle.ts:95`) — the retry hand-off loop and the "session registered but attempt failed" edge are the trickiest new code in the PR and deserve regression coverage per the repo's red-green policy. Relatedly, removing the `isRunning()` pre-check in `manager.ts` changes behavior for all platforms, not just DCM — it looks well-reasoned, but worth a sentence in the PR description.

On your kind offer to split `approvals` into its own PR: I'll leave that call to @anneschuth, but the bundling rationale (DCM needs an approvals story) reads as sound to me — the issues above are about gating consistency, not the bundling.

Thanks again for the care that went into this — especially the fourteen adversarial rounds and the honest limitations section. With item 1 fixed and item 2 addressed one way or the other, this is in very good shape.

---

Jadefalkner:
Thank you for the thorough trace — both blocking items were real, and both are now enforced rather than documented away (ffe95dd):

**1. `!approve` bypass — fixed.** `approvePendingPlan` now applies the same `resolveApprovals` participant gate as the reaction router, before touching the pending approval or the decision bridge. A platform-allowlisted non-participant gets a warning post and an audit log line, and the plan stays pending. Regression tests cover both the rejection and the participant-still-allowed path (red without the fix: 2 failures).

**2. `!invite` self-escalation — enforced, not documented.** `requireSessionOwner` now adds a participant requirement on top of the legacy check whenever the effective approvals mode is `owner`. That closes the self-invite hole and, by construction, the same class for every owner-gated command (`!kick`, `!cd`, `!permissions`, `!memory forget`, update commands) — a non-participant allowlisted user can no longer reshape a session they aren't part of. The legacy behavior is untouched when owner mode is not in effect, and invited guests gain nothing (the legacy owner-or-allowlist check still runs first). Documented in the Approvals section.

**Non-blocking items, all addressed:**
- *Message-based resume* is now owner-scoped in `message-handler.ts`, matching the reaction-based gate.
- *DCM distillation* skips explicitly with a debug log instead of failing silently — distilling from channel history would need a cross-platform channel-history API that doesn't exist yet; happy to build that as a follow-up if you want DCM sessions to feed channel memory.
- *`_inFlightSessionStarts` tests*: three regression tests — in-flight dedup delivering as follow-up, the retry hand-off loop (verified red against a single-check variant of the loop), and `sendFollowUp`'s bounded wait for a registered-but-not-yet-running session.
- *`isRunning()` pre-check*: added a paragraph to the PR description explaining why its removal is intentional for all platforms.

One test-infra observation from working on this: several `handler.test.ts` cases (`updateSessionHeader`, `kickUser`) hit their 5s budget intermittently when multiple test files run concurrently on a loaded machine — reproducible on this branch's *base* as well, so unrelated to this PR, but worth knowing if CI ever flakes there.

anneschuth:
Verified the new head (ffe95dd) — the security work all checks out:

- **`!approve` gate**: confirmed at the sink in `approvePendingPlan`, converging with the reaction router's gate; I found no third route to plan approval. Red-green reproduced exactly as you described (2 failures with the gates reverted, clean restored).
- **`requireSessionOwner` participant check**: confirmed, and the completeness trace holds — every way into `sessionAllowedUsers` (seed at start, `inviteUser`, message-approval invite decision) is now gated or participant-driven. The documented residual (allowlisted non-participants can still *message* under `owner` mode, but neither approve nor self-invite) reads as a reasonable, clearly-stated boundary.
- Claims 3–5 also verified; `bun test src/` is 2925/0 and typecheck is clean on the branch.

Two items still block a green merge, and one of them has a subtle root cause you can't see locally:

1. **The CI unit-test failure is still latent and will fail again on this head.** `resumeSession with direct channel mode › skips the thread-existence check for a synthetic DCM session id` (`lifecycle.test.ts:1677`) drives `resumeSessionImpl` all the way into `cli.ts` `start()`, which spawns the real `claude` binary — present on your machine (so the test "passes", by launching a live claude process), absent on GitHub runners (ENOENT → red). Reproducible locally with `CLAUDE_PATH=/nonexistent/claude bun test src/session/lifecycle.test.ts` → 1 fail. The neighboring "decision-bridge listener wiring" describe block (`lifecycle.test.ts:1527-1541`) already has the `CLAUDE_PATH` beforeEach/afterEach shim for exactly this — wrapping the DCM-resume describe the same way should do it.
2. **The integration failure** (`session-errors.test.ts:154`, "should handle session ending with simple response", `Expected: false` after a 500 ms sleep) wasn't touched. It may be a pre-existing timing flake, but since this PR changes session start/teardown timing (`_inFlightSessionStarts`), it deserves a look rather than a re-run-until-green — if it's the fixed sleep racing teardown, a bounded wait-for in the test would settle it either way.

One non-blocking suggestion: the message-based resume gate (claim 3) is the only one of the three security fixes without a regression test — worth adding while you're in there.

Once those are pushed and CI runs green, I'm happy to flip my review to approve.

---

Jadefalkner:
Both blockers and the test ask are in (ed986ea), on top of a rebase onto current `main` (Routines + 1.25.1 — CHANGELOG re-seated, the routines test block and the approvals-gate test block now coexist at the end of `handler.test.ts`):

1. **Latent CI failure — fixed with the shim you pointed at.** The DCM-resume describe now carries the same `CLAUDE_PATH` beforeEach/afterEach as the decision-bridge block. Verified with your exact repro: `CLAUDE_PATH=/nonexistent/claude bun test src/session/lifecycle.test.ts` → 1 fail without the shim, 91/91 with it. Nice catch on the silently-launched real CLI locally — that also means no more stray `claude` processes per local test run.
2. **Integration timing — bounded wait, and you were right about the mechanism.** The test slept a fixed 500ms and then asserted the session gone; teardown after the mock CLI's exit is asynchronous and #478's in-flight start tracking retimed it. It now uses the existing `waitForSessionEnded` helper (10s bound) that the neighboring test already used. Verified green against the Slack mock harness (`session-errors.test.ts`: 4/4).
3. **Message-based resume gate regression test** added in `message-handler.test.ts` (allowlisted non-participant under `approvals: owner` → no resume, warning posted). Red-green verified by temporarily reverting the gate condition: 1 fail without, 94/94 with.

Jadefalkner:
The red `integration (mattermost)` check turned out to be a third instance of the same disease (0ef0fc3): the `respondTo: mention` test waited for the first bot channel-root post and then *instantly* asserted the DCM session registered. The CI log shows the assertion failing at 07:54:37.04 while the mock CLI received the prompt at 07:54:37.18 — the post it found was the session header/sticky, which lands before `startSession` registers the session (registration is deliberately late for the in-flight dedup). The sibling describe in the same file already used a bounded `waitFor` for exactly this; the mention test now does too. #479 is rebased on top (552b3b8).

anneschuth:
Quick heads-up while I verify 0ef0fc3 (looking good so far — CI is green on both this and #479's head): #477 just merged, and its CHANGELOG entry under `## [Unreleased]` conflicts with this branch again. One more rebase over `main` when you get a chance — same trivial re-seat as before — and #479 on top. That should be the last one: nothing else is queued to touch main before this PR merges.

---

Jadefalkner:
Rebased over `main` with #477 merged — the CHANGELOG `[Unreleased]` section now carries both the DCM/approvals entries and #477's permission-prompt entry, nothing else changed. #479 re-stacked on top. Thanks for the merge!

anneschuth (approved):
Approving. Independently verified across the pushes:

- Both security gates (`!approve` participant gate at the sink, `requireSessionOwner` participant check) confirmed genuine, red-green reproduced, no bypass found — and confirmed intact after both rebases.
- The `CLAUDE_PATH` shim fixes the latent CI failure (verified with the repro: 91/91 with the shim, exactly-one-fail with it forced to a nonexistent path).
- Both integration deflakes are legitimate bounded waits with the original assertions retained — not test weakening.
- The message-based resume gate regression test is red-green against the real `handleMessage` path.
- The final rebase (2898316) drifts only in the CHANGELOG re-seat — zero code changes.
- On the branch: `bun test src/` 2992/0, typecheck clean, and CI + Integration Tests green on the current head.

Thank you for the exemplary follow-through on this one — the log-timestamp diagnosis of the third flake instance instead of a re-run was exactly right. Merging.

---

PR #479, opened 2026-08-19
feat: DM auto-discovery — cold direct messages spawn their own session (Mattermost)

Jadefalkner:
> Stacked on #478. Review/merge that one first; this adds one feature on top and can be taken or left independently.

### Motivation

The DCM PR gives you bot DMs the *static* way — one config entry per DM conversation. That is fine for one or two people and does not scale to a team: nobody wants to look up a DM channel id and edit config.yaml so a colleague can message the bot. This PR adds the *dynamic* way: anyone allowed can simply DM the bot "out of the cold".

### Two ways to DM the bot

| | Static entry (DCM PR) | Auto-discovery (this PR) |
|---|---|---|
| Setup | one `platforms:` entry per DM conversation, channel id looked up via API | `directMessages: true` on the main entry, nothing else |
| Who can DM | exactly the configured conversations | anyone on the entry's `allowedUsers` |
| First contact | must be configured before it works | just works — instance is spawned on the first message |
| Fits | personal setups, fixed 1:2 conversations | teams |

Both produce the same thing underneath: a direct-channel-mode session scoped to the DM partner.

```yaml
platforms:
  - id: mattermost-main
    type: mattermost
    # ... regular entry ...
    directMessages: true
```

### How it works

On the first DM from an allowed user, the bot derives a platform instance from the parent entry (`src/platform/dm-discovery.ts`): same credentials, `channelId` = the DM channel, `directChannelMode: true`, sticky hidden, `allowedUsers` scoped to the DM partner — which, with the DCM `approvals` default of `owner`, also scopes tool-permission prompts to that person. The first message is routed through the parent's connection while the instance's own connection comes up; afterwards the instance handles its channel itself.

Lifecycle is bounded: when the DM session ends (idle timeout, `!stop`), the derived instance and its connection are torn down. The next DM re-discovers the channel and **resumes the persisted session** — conversations survive both restarts and teardowns, and instances do not accumulate over uptime. Persisted DM sessions are also reconstructed at boot, before connect, so a bot restart resumes them directly.

Guard rails:

- First-parent-wins: with several `directMessages: true` entries on one bot account, exactly one instance owns a DM channel — the bot never double-replies.
- A failed connect tears the instance down instead of leaving a stale entry that would block the channel forever.
- A dedupe set covers the handover moment between parent-routed and instance-received messages.
- Group DMs (`channel_type: G`) are not discovered — only true 1:1 DMs.
- Empty `allowedUsers` means *everyone*, consistent with the rest of the bot — documented with a warning, since with `directMessages` that is every user who can DM the bot.
- A 2-second post-teardown sweep enforces the end-state invariant "no session outlives its platform" — whatever ordering the start/retry races take.
- Disabling a derived instance sticks: live discovery honors the toggle immediately, and after a restart the instance keeps a UI row (sourced from the enabled-state store) so it stays re-enablable.

### Why Mattermost only

Slack Socket Mode distributes event envelopes across an app's active connections; per-DM connections on shared credentials would swallow each other's events. Slack DM support needs a single-connection, channel-aware implementation — out of scope here and called out in the docs.

### Known timing windows (open points, stated upfront)

Discovery is inherently concurrent (parent socket, per-instance socket, grace/reaper timers), and I would rather declare the residual windows than claim there are none. What can happen in the worst case:

- A message arriving in the seconds-wide handover between parent-routed delivery and the instance's own connection can be **duplicated** (Claude answers twice) or — after an instance teardown — **dropped**: the user sees no reply, writes again, and the next message re-discovers the channel and resumes the persisted session.
- A session stranded by a failed connect can appear **frozen** until the idle timeout persists it; the next DM resumes it.
- If a reaper edge is missed, an idle instance's WebSocket lingers until the next bot restart.

What can **not** happen: no timing window bypasses an authorization gate (allowlist, partner scoping, `approvals`, group-DM exclusion are all checked synchronously on the message path), no persisted conversation is lost, and no state survives incorrectly across a restart. In short: at-least-once chat semantics with self-healing, no security or data-loss paths.

One open point remains:

- The handover dedupe uses a bounded LRU of routed post ids; a burst larger than the window could theoretically re-admit a duplicate.

### Architecture & testing

The whole lifecycle lives in an injectable runtime (`src/platform/dm-discovery-runtime.ts`, `createDmDiscoveryRuntime(deps)`) — `index.ts` only wires it. That makes the exact production code path testable:

- **20 unit tests** pin the invariants: allowlist gating, connect-window routing and its dedupe, connect-failure teardown (including cancelling a session stranded on a dead client), grace-period teardown and its cancellation by a new session, the stale-timer ABA guard, orphan reaping (sparing resumable persisted conversations), boot-reconstruction edge cases, and generation-bound settling. Plus the id-scheme/config-derivation tests.
- **A real-Mattermost integration suite** exercises the production runtime end-to-end via the test harness: a cold DM spawns the instance and the bot replies in the DM at channel root; follow-ups reuse the session; a DM session and a channel session coexist independently.

Also live-verified in production use: cold first contact answers within a second, restarts resume DM conversations, teardown + re-discovery resumes the same conversation. Hardened through several dedicated adversarial review rounds (races, teardown, multi-parent, reconstruction edge cases).

---


anneschuth (commented):
Thank you for this one too — and particularly for the "known timing windows" section. Declaring the residual races up front instead of claiming there are none is exactly the right way to present a change like this, and it held up: I traced the authorization paths and agree with your central claim. The allowlist check happens synchronously before anything else (`dm-discovery-runtime.ts:146`, including the connect-window forwarding path), the derived instance re-checks against the partner-scoped `allowedUsers`, group DMs fail closed on missing/non-`D` `channel_type`, and first-parent-wins registration has no await between check and set. The unit tests genuinely exercise the hard cases (stale grace timer vs. replacement client, retry-generation waits, boot-settle guards), and the injectable runtime is the real production module — nice.

Since this is stacked on #478 (which has a review with two blocking items), just a handful of asks to fold in when that settles:

1. **Stale TUI rows on teardown**: `index.ts` adds a `ui.setPlatformStatus` row per discovered DM, but `teardown()` never removes it — every DM channel ever contacted leaves a permanent row shown as enabled after the instance is gone. Bounded (one per distinct channel) but misleading.
2. **Bound the `_inFlightSessionStarts` wait**: the `for (;;)` loop in the connect-failure path (`dm-discovery-runtime.ts:195`) awaits an underscore-private map from `src/session/lifecycle.ts` with no timeout. If a start promise ever fails to settle, the dead instance owns that channel forever and blocks rediscovery. A timeout (or a public lifecycle API instead of the underscore import) would remove the failure mode.
3. **`parseDmPlatformId` appears unused in production** — exported but only referenced by its own test (reconstruction uses config-prefix matching instead). Either wire it in or drop it; knip may flag it.
4. **Small asymmetry worth a comment**: boot reconstruction widens the derived platform's `allowedUsers` to `persisted.sessionAllowedUsers`, while live discovery scopes to the partner alone. Practically harmless in a 1:1 DM, but a comment explaining the intent would prevent someone "fixing" it either way later.

One design question, genuinely open rather than a demand: the nine hardening-round commits suggest the per-DM websocket model is inherently race-prone territory — the parent's connection already receives every event (that's how discovery works), so routing DM posts through the parent permanently would eliminate the handover window and most of the grace-timer/reaper/generation machinery. The current design is defensible (it maximally reuses the per-platform machinery, including MCP permission-server scoping), so this is not blocking — but if you considered and rejected parent-socket routing, I'd love a short note on why, for the archaeology.

The declared duplicate/dropped-message windows read as acceptable UX artifacts to me given the self-healing behavior. Overall this is careful work — looking forward to it landing once #478 is settled.

---

Jadefalkner:
All four asks are in (e7d6058, rebased on the updated #478):

1. **Stale TUI rows** — `teardown()` now removes the instance's status row via a new `removePlatformStatus` on the UI provider API (wired through headless + ink providers). Deliberately spared: rows of *disabled* instances, which are the only re-enable handle — covered by two new runtime tests (removed on plain teardown, kept when disabled).
2. **Bounded in-flight wait** — the connect-failure loop now polls with a 30s deadline (`Promise.race` against a 1s tick) and proceeds to teardown with a warning if the start promise never settles. The existing end-state sweep already covers a session that registers after we move on, so the failure mode reduces to "one message lost", within the declared model. I kept the underscore import for now — a public lifecycle API felt like #478 surface; happy to lift it there if you prefer.
3. **`parseDmPlatformId`** — dropped, along with its tests. Digging into why reconstruction didn't use it turned up a real reason to keep it gone: a parent platform id may itself contain the `--dm-` separator, so parsing at the first occurrence is ambiguous — longest-prefix matching against the configured parents is the only correct resolution. Left a NOTE at the former definition site so it doesn't get reintroduced.
4. **`allowedUsers` asymmetry** — commented at the reconstruction site: live discovery scopes to the DM partner (the only counterparty a fresh 1:1 DM can have), boot reconstruction restores the persisted participant set because an `!invite`d collaborator must survive a restart. Explicitly marked "do not fix either side to match the other".

**On the design question — parent-socket routing:** considered, and rejected for one load-bearing reason plus one economic one. The load-bearing one: the MCP permission server scopes by platform instance — it opens its own connection with the derived instance's credentials/channel and listens for reactions there. Routing DM posts through the parent permanently would keep *message delivery* simple but leave *approvals* needing a per-DM identity anyway (or channel-aware filtering inside the MCP server, changing its scoping assumptions for every platform). Once the derived instance must exist for approvals, giving it its own socket buys the entire per-platform machinery (reconnect/backoff, sticky/header suppression, session persistence keying) for free. The economic one: the handover window machinery is ~contained in the runtime module, while parent-routing would smear channel-awareness across MCP server, reaction router, and delivery paths. That said — if the per-DM socket count ever becomes a scaling problem (it's one websocket per active DM conversation), parent-routing with an MCP redesign is the right successor, and I'd rather do it as its own PR than fold it in here. Happy to add this rationale to the module docblock if you want it in-tree.

anneschuth:
Verified e7d6058 — all four asks are implemented correctly: the teardown UI-row removal with the deliberate disabled-instance carve-out (nice touch keeping the re-enable handle), the 30s-bounded wait (checked the race logic — no busy-loop, no unhandled rejection, and the end-state sweep covers the late-registration case), the `parseDmPlatformId` removal (the separator-ambiguity finding is a genuinely good reason), and the asymmetry comment. 24/24 DM-discovery tests and typecheck are green on the branch, and the delta contains nothing beyond the claimed fixes. The parent-socket rationale is convincing — the MCP scoping argument settles it; a short version in the module docblock would be welcome but isn't blocking.

From this PR's side there's nothing left to ask. It remains gated on #478, including the `CLAUDE_PATH` test fix noted there (this branch inherits that test verbatim, so its CI stays red until that lands and the rebase picks it up).

---

Jadefalkner:
Rebased onto the updated #478 head (80864f2) — only the stack moved, no content changes on top of e7d6058's review fixes. DM-discovery suite 24/24, typecheck clean.

anneschuth:
#478 is merged (squash, `1dbb14a`) — thank you again for the stack. The squash rewrite means this branch's copies of the #478 commits now conflict with main (5 files: CHANGELOG, docs/CONFIGURATION.md, src/index.ts, src/onboarding.ts, tests/integration/helpers/bot-starter.ts). A rebase that drops your #478 commits should resolve it cleanly:

```
git fetch upstream main
git rebase --onto upstream/main <last-478-commit> feat/dm-auto-discovery
```

Content-wise this PR is fully verified and approved on my side — once the rebased head is up and CI is green, it merges straight away. Apologies for the churn; this is the last rebase in the pipeline (nothing else touches main before this lands).

---

Jadefalkner:
Rebased with exactly your recipe — the #478 commits are dropped, only the twelve DM-discovery commits remain on top of `main` at `1dbb14a`. DM-discovery suite 24/24 and typecheck clean locally; CI should confirm. Thanks for the smooth pipeline — and for merging the stack so quickly!

Jadefalkner:
The red `integration (mattermost)` on ca6fe6b was the decision-bridge question test timing out on the runner (`Session Questions > should deliver the answer through the decision bridge`, 20s wait for the confirmation post) — a file this PR doesn't touch, and green on #478's identical base an hour earlier. I can't re-run jobs on this repo, so I retriggered CI with a no-op amend (bdb4c76, identical tree).

PR #480, opened 2026-08-19
fix: tie keep-alive inhibitors to the bot process lifetime

Jadefalkner:
## Problem

The keep-alive child processes only die on a clean teardown (`sessionEnded()` / `forceStop()`). When the bot process dies hard — SIGKILL, OOM, or a crashed test runner — the `systemd-inhibit sleep infinity` child is reparented to init and keeps blocking system sleep **forever**.

This is not hypothetical: after a day of running the integration suite locally, `systemctl hibernate` was blocked by **43 orphaned inhibitors**, all reading "Claude Code session active", all with ppid 1.

## Fix

Tie the inhibitor's lifetime to the bot process at the OS level, so no teardown path is required:

- **Linux**: `systemd-inhibit` now execs `cat` reading from a pipe held by the bot instead of `sleep infinity`. When the bot dies (even SIGKILL), the kernel closes the pipe, `cat` exits on EOF and the lock is released. Event-driven, no polling.
- **macOS**: `caffeinate -s -i -w <botpid>` — caffeinate exits natively when the watched pid dies.
- **Fallbacks** (xdg-screensaver loop, PowerShell `SetThreadExecutionState`): the loop polls the parent pid and exits when it is gone.

Spawn specs are extracted into pure helpers (`keepAliveSpawnSpec`, `linuxFallbackScript`, `windowsScript`) so the lifetime coupling is unit-testable.

## Verification

- New regression tests assert the coupling per platform (red without the fix, green with it).
- Manual proof on Linux (NixOS, systemd 256): started a KeepAliveManager in a child bun process → 1 inhibitor lock active → `kill -9` the child → lock gone within a second. Before this change the lock survived the kill.
- `bun test`, `tsc --noEmit` and `eslint` clean (the pre-existing `setMode` SUID / `validateOutboundPath` failures on this machine are unrelated and fail identically on `main`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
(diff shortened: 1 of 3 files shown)


anneschuth:
Thanks!

PR #484, opened 2026-08-20
test-infra: deflake handler tests — mock environment probes, raise the per-test budget

Jadefalkner:
## Problem

Several `handler.test.ts` cases (`updateSessionHeader`, `kickUser`, `setRespondOnlyWhenMentioned`) die intermittently at exactly the 5s per-test budget when other processes compete for the machine — local pre-commit hooks running alongside a test batch, or a loaded CI runner. In isolation the same tests pass in milliseconds. First noticed while working on #478.

Reliable repro: run `bun test src/platform src/claude` and `tsc --noEmit` concurrently with the handler suite — one of the header-refreshing tests then hits the budget on most attempts.

## Root cause (two layers)

1. **Unit tests were probing the real environment.** `updateSessionHeader` — reached by most command handlers — calls `isGitRepository`/`getCurrentBranch` (each spawns a `git` child process against the mock session's nonexistent `workingDir`) and `formatBatteryStatus` (sysfs reads on Linux, a `pmset` exec on macOS). Under fork/exec contention those spawns stall for seconds. None of the affected tests assert git or battery rows.
2. **The 5s default budget has no headroom for event-loop contention.** Even with the spawns mocked, a fully saturated machine can occasionally stretch a millisecond test past 5s.

## Fix

- `handler.test.ts` mocks the environment probes (`git/worktree.js`, `utils/battery.js`) via `mock.module`, with the module under test imported afterwards. **The `afterAll` restore stores a pre-mock *value snapshot* (`{ ...namespace }`), not the live namespace** — bun module namespaces are live bindings, so after `mock.module` the saved namespace itself yields the mocks and a namespace "restore" restores the mocks (verified with a minimal two-file experiment on Bun 1.3.13). The existing `crossSpawn` restore in `plugin/handler.test.ts` had the same hazard and is fixed the same way.
- The per-test budget is raised to 15s via `--timeout 15000` on every unit-test entry point: `package.json` scripts, lint-staged, `ci.yml`, `release.yml`, plus the documented invocations in `CONTRIBUTING.md`, `CLAUDE.md` and the PR template (now `bun run test`). Note: bunfig's `[test]` section does not support a timeout key — verified empirically with a 6s dummy test that still timed out at 5s.
- On the number: 15s is a deliberate trade. It is high enough that load spikes stop producing false reds, and a genuinely hung test still fails fast. The cost — a real 6–15s slowdown regression would no longer fail the suite — seems acceptable for a codebase whose slowest legitimate unit test runs in tens of milliseconds; if you'd rather keep a tighter global budget with per-test overrides on the header suites only, happy to switch.

## Verification

- Repro recipe before: flaked in most rounds (budget kill at ~5–7s, failures wandering between header-refreshing tests). After the probe mocks: sharply reduced but one residual hit (pure event-loop contention). After both changes: 10/10 stress rounds green.
- Cross-file restore: minimal experiment confirms the snapshot yields the real implementations to later test files while the live-namespace pattern yields the poisoned bindings.
- `tsc --noEmit`, eslint, and the affected suites clean; branch is rebased onto current `main` (post-Routines).


anneschuth (approved):
Approving — verified in depth, and this held up impressively:

- **The live-binding claim is real and your fix is load-bearing.** Reproduced empirically: after `mock.module`, a pre-mock namespace reference yields the mock, while the `{...ns}` value snapshot retains the real implementations — so the `plugin/handler.test.ts` `crossSpawn` restore was indeed a latent bug, and the `afterAll` restore genuinely protects later files in the same process (confirmed `battery.test.ts` / `worktree/handler.test.ts` see the real implementations in an ordered run: 108/0).
- **Entry-point sweep is complete**: package.json scripts, lint-staged, ci.yml, release.yml, and the docs all updated; no stale 5s invocation remains. Also reproduced your bunfig finding — the `[test]` timeout key is silently ignored — so the CLI flag everywhere is the only option.
- Full suite at the head: 2959/0, typecheck clean.
- On global-15s vs per-test overrides: your mocks fix the root cause, so I'm fine with the global budget as belt-and-suspenders.

Two optional nits, take or leave before merge:
1. A CHANGELOG `[Unreleased]` entry (repo convention — a short "Changed" note on the test budget is enough).
2. The "import AFTER mock.module" comment at `handler.test.ts:34` slightly overstates the mechanism — the static import is hoisted; it works because Bun retroactively patches live bindings. A one-line rewording would spare the next reader the confusion.

**Merge sequencing:** this lands right after #479 (I promised that branch no further rebase churn, and these two overlap on docs). No action needed from you for that — I'll merge when the queue clears.

---

Jadefalkner:
Both nits taken (10ee7d7, rebased onto current `main` so the CHANGELOG lands under the existing `[Unreleased]`):

1. **CHANGELOG** — a `### Changed` entry covering the probe mocks, the live-binding snapshot restore, and the budget raise.
2. **Comment reworded** — you're right that it overstated the ordering: hoisted static imports (`MemoryStore` → `git/worktree`) load the modules before `mock.module` runs, and it works because Bun retroactively patches live bindings of already-loaded modules. The comment now says exactly that; the dynamic import stays purely for clarity of intent.

PR #487, opened 2026-08-20
feat: ackReaction — read-receipt reaction on accepted messages

Jadefalkner:
## What

Opt-in per-platform option `ackReaction`: the bot reacts to every message it **accepts for processing** — session start, follow-up in an active session, resume of a paused session — the instant it is accepted, before Claude produces any output. `true` uses 👀 (`eyes`); a string names a custom emoji. Default off, so existing setups are unchanged.

```yaml
platforms:
  - id: mattermost-main
    type: mattermost
    ackReaction: true          # 👀 on every accepted message
    # ackReaction: white_check_mark   # or any custom emoji name
```

## Why

The typing indicator is transient and platform-dependent; in a busy channel (or DCM, where the channel is the session) there is no durable signal that a message actually reached the bot. The gap is worst for messages that arrive during an in-flight session start: they are correctly queued and delivered as follow-ups (#478), but the user sees nothing at all until Claude's first output. A persistent reaction closes that gap the way chat-native read receipts do.

The pattern is borrowed from a private Mattermost bot that has used an immediate 👀 ack for months — it turns "did the bot see this?" into a non-question.

## Design notes

- The ack fires **only after all accept gates**: mention/quiet-mode checks, allowlist, session-participant checks. Ignored side conversations and unauthorized messages are never acked (no information leak about the allowlist to a rejected sender beyond the existing warning posts).
- Fire-and-forget: a failed `addReaction` is swallowed — a missing receipt must never block or fail message handling.
- Plumbing follows the `approvals` option exactly: config type → `PlatformClient` field → both clients → resolver in `platform/utils.ts`.

## Semantics and limits

- The receipt means **accepted, not delivered**: a later failure (capacity limit, Claude not coming up) is still reported by its own post. `!commands` are not acked — they have their own immediate feedback. In DCM every accepted message costs one reaction API call. All documented in the option's docs row.
- Malformed config values (`ackReaction: 123`) warn once at client construction and disable the feature instead of silently enabling the default emoji. Failed reactions (typically a nonexistent custom emoji name) log at debug level for diagnosability.

## Verification

- 7 new tests in `message-handler.test.ts` (default-off, session-start ack, custom emoji, follow-up ack, resume ack, no ack on ignored messages, no ack for unauthorized users) plus normalize/resolve unit tests in `platform/utils.test.ts`. Red-green verified by inverting the resolver guard: 4 fail without the feature wiring, all green with it.
- `tsc --noEmit` and eslint clean; docs table rows (both platform tables) and CHANGELOG entry included.


anneschuth (approved):
Approving — verified in depth:

- **Gate placement is correct on all three paths** (follow-up after the side-conversation/quiet-mode/`isUserAllowedInSession` gates, resume after the owner-scoped auth gate, new session after mention/allowlist/empty-prompt/command gates, DCM included). No path acks an ignored or unauthorized message — no allowlist information leak.
- Fire-and-forget is genuinely non-blocking with no unhandled-rejection path.
- Red-green independently reproduced on the unauthorized-no-ack test (ack moved ahead of the gate → exactly that test fails).
- `bun test src/` 3031/0, typecheck and lint clean, all 10 CI checks green on the head. The pre-emptive hardening round (malformed-config warn-and-disable, debug-logged failures) is appreciated.

Four optional follow-ups, none blocking — take them in a later PR if you like:
1. The message-approval flow (non-participant's message approved by a participant via 👍) is a fourth "accepted" avenue that never gets the receipt — ack it or note the exclusion in the docs.
2. Mattermost/Slack asymmetry: `ackReaction: "👀"` (literal emoji) works on Slack (`getEmojiName` conversion) but silently no-ops on Mattermost; normalizing in `normalizeAckReaction` would close it.
3. A dedicated no-ack test for the follow-up path's `isUserAllowedInSession` gate (only the new-session gate is red-verified).
4. Two stray blank lines in `src/message-handler.test.ts`.

Merging. Thanks for another clean one!

---

PR #488, opened 2026-08-20
feat: auditLog — append-only per-platform audit trail

Jadefalkner:
## What

Opt-in per-platform option `auditLog: true`: one append-only JSONL stream per platform under `~/.claude-threads/audit/<platformId>.jsonl` (override: `CLAUDE_THREADS_AUDIT_DIR`) recording what the bot did:

- `tool_use` — every tool call Claude **issued**, incl. `server_tool_use` and subagent sidechains (marked `subagent: true`), with the audit-relevant detail: Bash command line, file path, search pattern (capped at 500 chars).
- `session_start` / `session_resume` / `session_end` — full lifecycle incl. failure paths (`start-failed`, `resume-failed`, `resume-exit:<code>`, `timeout`, `pause`, `exit:<code>`, `kill`), exactly one end per session via an idempotency guard.
- `command` — security-relevant `!commands` with actor: `!cd`, `!invite`, `!kick`, `!permissions`, `!stop` (active and paused), `!kill`, `!memory forget`, `!routines` management, routine creation, `!worktree remove`, `!plugin install`/`uninstall`.
- `plan_approval` — approved/denied, by whom, via reaction or `!approve`.

```yaml
platforms:
  - id: mattermost-main
    type: mattermost
    auditLog: true
```

## Why

Thread logs are debugging artifacts: per-thread, raw events, deleted after 30 days. An audit trail is the opposite: one flat greppable stream per platform, spanning sessions, that the bot **never deletes** — built for SIEM file ingestion (Wazuh/Splunk file collectors read JSONL natively). For a multi-user bot the question "who ran what, when, in which thread" deserves a first-class answer.

## Design notes

- **Semantics: the trail records the request** at the moment Claude emits it. An interactive permission denial can still stop the execution — and the denied attempt is exactly what an auditor wants to see. Per-tool allow/deny decisions themselves are out of scope: they resolve inside the MCP permission server subprocess, which the bot process does not observe (documented; plan approvals and audited commands cover the decisions that flow through the bot).
- **Writes are synchronous on a cached fd** — one `write(2)` per entry, no open/close per write. Deliberately unbuffered: a trail that loses its tail on crash defeats the purpose, and the volume (tool calls, commands) stays far below where batching would matter.
- **Hardened against leftover artifacts**: `0600`/`0700` enforced via `fchmod`/`chmod` even on pre-existing files and directories (creation modes alone silently reuse a world-readable leftover), the writer refuses symlinked audit paths (`lstat` + `O_NOFOLLOW`), filenames are collision-free via `encodeURIComponent`.
- **Actor attribution is best effort and documented**: the username whose authorized message triggered the current turn (resumes attribute to the resuming user; attribution is assigned only after the authorization gate), falling back to the session starter.
- Derived DM platforms inherit the parent's `auditLog` and clean up their registry entry on teardown.

## Verification

- 15 new tests: module behavior (JSONL shape, permissions incl. pre-existing-file re-tightening, symlink refusal, detail cap, collision-free encoding, enable/disable), event-tap coverage (tool_use, subagent marking, disabled no-op), and the `!kill` command tap (recorded for authorized, absent for unauthorized). Red-green verified per tap by temporarily removing the wiring.
- `tsc --noEmit`, eslint, and all affected suites (402 tests) clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+- **Audit trail (`auditLog`)** - Opt-in per platform: an append-only JSONL stream per platform (`~/.claude-threads/audit/`, files `0600` enforced even on pre-existing artifacts, symlink-refusing writer, never deleted by the bot) recording what the bot did — every tool call Claude issued incl. `server_tool_use` and subagent sidechains (with Bash command line / file path / pattern as detail), session lifecycle incl. failure paths with the triggering user, security-relevant `!commands` (`!kill` and paused-session `!stop` included), routine creation, worktree/plugin mutations, and plan approvals with decider. Built for SIEM file ingestion; rotation/retention is the operator's call. Tool-permission allow/deny decisions stay out of scope (they resolve inside the MCP permission server subprocess; the issued request is still recorded).
+
 ## [1.27.0] - 2026-08-20
 
 ### Added
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -205,7 +205,7 @@ The approval set is fixed when the Claude CLI is spawned; a later `!invite` exte
 
 Notes for operators:
 
-- **The bot never deletes audit files.** An audit trail that expires itself is not one — rotation and retention are yours (logrotate, or let your SIEM's file collector ingest and rotate).
+- **The bot never deletes audit files.** An audit trail that expires itself is not one — rotation and retention are yours (logrotate, or let your SIEM's file collector ingest and rotate). The writer holds the file descriptor open across writes, so use **`copytruncate`** (or restart the bot after rotating): a rename-based rotate never errors the cached fd, and the bot would keep appending to the rotated file until restart.
 - **Entries contain command lines verbatim** (that is the point); files are `0600` in a `0700` directory — enforced on every start, including pre-existing files, and the writer refuses symlinked audit paths. Treat the directory with the same care as the thread logs.
 - **Actor attribution is best effort**: the username whose (authorized) message triggered the current turn — resumes are attributed to the resuming user — falling back to the session starter. In fast multi-user threads a tool call can be attributed to the previous sender.
 - **Tool-permission decisions (allow/deny of individual tool calls) are not recorded** — they are resolved inside the MCP permission server subprocess, which the bot process does not observe. Plan approvals and the audited commands cover the decisions that flow through the bot itself.
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -132,6 +132,7 @@ stickyMessage:
 | `directChannelMode` | No | Direct channel mode: the whole channel is one session, and the bot replies with top-level channel posts instead of thread replies. `true` for defaults, or an options object (`respondTo`). See [Direct Channel Mode](#direct-channel-mode). |
 | `approvals` | No | Who may answer tool-permission prompts and other reaction gates: `owner` (session participants) or `all_users` (everyone on `allowedUsers`). Unset keeps the historical default per mode — `all_users` for thread sessions, `owner` for direct channel mode. See [Approvals](#approvals). |
 | `ackReaction` | No | Read receipt: react to every accepted message (session start, follow-up, resume) the instant it is accepted, before Claude produces output. `true` uses 👀 (`eyes`), a string names a custom emoji. Persistent, unlike the typing indicator — useful in busy channels and for messages queued behind an in-flight session start. The receipt means *accepted*, not *delivered*: a later failure (capacity limit, Claude not coming up) is still reported by its own post. `!commands` are not acked — they have their own immediate feedback. Note: in direct channel mode this is one reaction API call per accepted message. Default off. |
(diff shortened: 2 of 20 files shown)


anneschuth (changes requested):
This is careful work, and the security core verified out fully — I want to lead with that:

- **The hardening claims are all real.** `openAuditFd` re-tightens pre-existing files and dirs (unconditional `chmodSync`/`fchmodSync`, not just creation modes), the `lstat` + `O_NOFOLLOW` pair closes the symlink TOCTOU for the final component (and the parent is bot-created 0700, so the parent path is trusted), `encodeURIComponent` is genuinely collision-free. I red-checked both security tests myself: stripping the chmods fails the re-tightening test, stripping the symlink protections fails the refusal test — both exercise the real code path.
- Error containment in the writer is correct (an ENOSPC or deleted dir can never break message handling; fd dropped and retried), the 500-char cap is applied after the spread so callers can't bypass it, secret exposure is by-design and honestly documented, and **actor attribution is correctly gated** — `lastActorUsername` is only assigned after the fail-closed auth gate, every audited command sits inside its own auth check, and an unauthorized user cannot poison attribution.
- 1687 tests / 0 fail on the affected areas, typecheck and lint clean. One-write-per-tool-call verified against the real-CLI captures — the tap is per-event, not in the streaming path.

**One correctness bug blocks the merge — the `auditCause` parameter is dead:**

`closeThreadLogger(session, 'kill', ...)` at `lifecycle.ts:2206` runs before `removeFromRegistry(session, ctx, auditCause)` at :2222, and the exactly-one-end guard means the first writer wins — so `cleanupIdleSessions`' `killSession(..., 'timeout')` (:2312) records `detail: 'kill'`, and the `'timeout'` never appears in the trail. Same at :2146/:2170: `closeThreadLogger(session, 'exit')` precedes the `exit:${code}` qualification, so a non-zero exit code is unreachable. For an audit log, "timed out" vs "killed by a user" is exactly the distinction that matters, and the PR description promises it ("session lifecycle incl. failure paths"). Fix looks small: pass the cause into `closeThreadLogger`/`cleanupSession`, or record `auditSessionEnd(session, cause)` at the top of `killSession` before the logger closes. Please add a regression test asserting the recorded cause (none of the lifecycle tests currently assert it, which is how this slipped through) — the timeout path (`'timeout'` in the trail, not `'kill'`) is the important one.

Two optional nits, not blocking:
1. The tool_use tap in `events/handler.ts:179-207` reads `block.type` outside a try/catch; a pathological event shape would throw past the audit tap and skip MessageManager handling for that event. Wrapping the tap body would make the "never breaks message handling" property airtight.
2. The rotation docs say "logrotate" generically, but a rename-based rotate never errors the cached fd, so the bot appends to the rotated file until restart — worth a `copytruncate` note.

Also heads-up: #487 just merged, so this branch will need the usual CHANGELOG re-seat over `main` along with the fix.

With the cause fix + test in, this is ready — the hardest parts are already right.

---

Jadefalkner:
Thanks for the thorough verification — the dead `auditCause` is fixed, and chasing it surfaced three more paths with the same first-writer problem, so the fix covers all of them:

- **`closeThreadLogger` takes an `auditReason` override** (the audit cause can be more specific than the logger action), and `cleanupSession` passes it through. The kill site records the caller's cause — so `cleanupIdleSessions`' `'timeout'` now actually lands — the interrupt-pause site records `'pause'`, the shutdown site `'shutdown'`, and the pre-response exit `'early-exit'`.
- **Non-zero exits now record `exit:<code>` for real sessions.** Your review was right that the code qualification was unreachable — and it turned out the *other* branch was the one real active sessions hit: `wasResumed` is `state !== 'starting'`, so every active session's non-zero exit recorded `resume-exit:<code>` whether it was ever resumed or not. That branch now records the plain fact, `exit:<code>` (`exit` for a signal death, matching the clean-exit path); genuine resumes are still identifiable from `session_resume` entries in the same stream, and `resume-failed` is untouched.
- **Emergency `!kill` (`killAllSessions`) records the cause per session** before the raw kill — previously those sessions fell through to whatever the generic exit handlers happened to write, or nothing.

The regression tests assert the recorded cause on all eight paths (timeout, kill, killAll, pause, exit 0/2/null, shutdown, early-exit) with exactly one `session_end` per scenario, and each fix was red-verified — reverting it fails exactly its tests. The active-session fixtures mirror production state (`markClaudeResponded` → `active`), which is how the unreachable-branch test from the first draft got caught and replaced.

Both nits are in: the audit tap is wrapped in try/catch so a pathological event shape can't skip message handling, and the rotation docs now call out `copytruncate` (the writer holds the fd open across writes). Rebased over `main` with the CHANGELOG re-seated under `[Unreleased]`.

anneschuth (approved):
Approving — the fix push verified fully:

- **The cause plumbing is real and complete**: `closeThreadLogger`'s `auditReason` override reaches every teardown site (`timeout`, `pause`, `shutdown`, `early-exit`, caller's cause on kill), `killAllSessions` records per session before the raw kill, and the once-guard still holds (no double `session_end` anywhere).
- **The `wasResumed` relabel is correct** — your analysis matched mine: every active session's non-zero exit was landing in the resume branch; it now records plain `exit:<code>`/`exit`, with `resume-failed`/`start-failed` untouched. Good catch chasing the original finding into the neighboring paths.
- **Nine regression tests** assert the exact recorded cause with exactly-one-end per scenario. I red-checked two of the fixes independently: reverting the override plumbing fails exactly the four dependent tests; reverting the exit relabel fails exactly the exit:2 and signal tests.
- The security module is byte-identical to the previously verified version, both nits are in, the rebase is clean, full suite 3056/0, typecheck clean, and both CI workflows green on the head.

One cosmetic note, take or leave: the backstop at `lifecycle.ts:2182` would format a null code as `exit:null`, but it's unreachable (the logger always writes first) — fine as is.

Excellent work across both rounds. Merging.

---

PR #497, opened 2026-08-24
Agent tools: Claude can use memory, routines and watches from inside a session

anneschuth:
## Summary

Six new MCP tools let Claude itself use the bot's features mid-session: `remember_fact`, `list_memory`, `propose_routine`, `propose_watch`, `list_routines`, `list_watches`. The design keeps every write path either visible or human-gated, and went through **three adversarial review rounds** (7 → 8 → 6 findings, severity declining to nil; all verified findings fixed, each behavioral fix red-green tested).

## Architecture

The stores, their per-platform locks, caps, and the audit log all live in the **bot process**, while the session's MCP server is a separate child. Tool calls travel over the session's existing **decision bridge** as a new `agent_action` request kind and execute in `src/operations/agent-actions/handler.ts` — where every gate is authoritative. The env flags that decide which tools the MCP child registers (`src/mcp/agent-features-env.ts`) are advisory tool-list hygiene, re-checked bot-side per call.

`ClaudeCliOptions.agentFeatures` is **required** (like `memory`) so no spawn/respawn path can silently drop or widen the gates — all five respawn sites carry it through `buildRestartCliOptions`.

## Gating matrix

| Tool | Gate |
|---|---|
| `remember_fact` | No approval prompt (distiller precedent) — instead: **visible thread post** naming the undo path, audit log, **5 writes per session** (slot reserved synchronously so a parallel burst can't race past it; duplicates/errors refund), over-cap text refused (never truncated), `agent` provenance in `!memory`, and agent entries can **never supersede a user's entry** (they evict with distilled ones). Refused in **unattended sessions** — a prompt-injected routine/watch fire must not seed future sessions' context. |
| `list_memory` / `list_routines` / `list_watches` | Read-only, config-gated. `list_memory` keeps the *newest* 100 with `!memory forget <n>`-aligned indices. |
| `propose_routine` / `propose_watch` | **Save nothing.** They post the existing confirmation card — badged "Claude proposes…", with model-authored text collapsed to a single line so it can't restyle the card — and return `proposed_awaiting_human_approval`. The **decision is owner-gated** (an `!invite`d guest can neither approve *nor veto* — unauthorized reactions are refused *without consuming the proposal*, warned once). A proposal never displaces a pending human confirmation. `createdBy` = session owner (per-fire re-authorization depends on it). Refused in DCM (tools not even registered there) and in **unattended sessions**. |
| forget / pause / resume / delete / manual run | **Not exposed** to the agent at all. |

**Loop prevention:** routine/watch-fired sessions are marked `unattended` (on `Session`, persisted, resumed intact — with a prompt-prefix heuristic so sessions persisted by a *pre-upgrade* bot fail closed, not open). Such sessions can list, but cannot write memory or propose.

## Review rounds (all findings fixed unless noted)

- **R1:** approval owner-gate at save time; unattended memory-write gate; cap slots only for actual saves; announcement-failure containment; `!memory` `agent` label; bridge deny-shape mapping; `proposedByAgent` audit attribution.
- **R2:** synchronous cap reservation (parallel-burst race); guest reactions no longer consume/veto proposals (pre-consumption executor gate); fail-closed resume heuristic for pre-upgrade unattended sessions; `list_memory` newest-tail; refuse-over-truncate for propose fields; abort-signal check in `remember_fact`; `CT_DCM` registration gate; shared `entrySourceLabel`.
- **R3:** single-line collapse of model-authored card text (card-injection guard); pending-slot protection; warn-once spam guard; honest system-prompt wording for unattended tool absence; worktree ops dedup. Documented residual: an abort *during* the awaited platform call can still land the (valid, human-decidable) side effect — bounded by the 15s client timeout vs sub-second posts.

## Testing

- 48 new unit tests across the feature + review rounds, red-green verified for every security gate (unattended propose/memory, supersede authorization, approval owner-gate, cap race burst, pre-consumption guest gate).
- Full battery per push: typecheck, lint, knip, build, `bun test src/` (3205 pass / 0 fail), Slack integration suite (136 pass / 20 platform skips).

## Docs

`docs/MCP-TOOLS.md` (per-tool reference incl. guardrails), `docs/CONFIGURATION.md`, `CLAUDE.md` (invariants), CHANGELOG.

## Checklist

- [x] Tests pass (`bun test`)
- [x] Documentation updated (if needed)

https://claude.ai/code/session_012oYWi6VCBjsJdrK6CBeNe1


anneschuth (commented):
Deep review with independent verification — **the security architecture holds up exactly as described**, and this is ready to merge once one trivial CHANGELOG conflict is resolved. (Posted as a comment since the token can't approve its owner's PR.)

## What was verified independently

Ran the full suite on the PR head: `tsc --noEmit` clean, `bun run lint` clean, **3205/3205 tests pass**. Beyond that, four load-bearing gates were red-green checked (fix removed → targeted tests must fail, then restored):

1. **Unattended gates** stripped from `agent-actions/handler.ts` → 2 tests fail ✓
2. **`canSupersede`** weakened in `memory/store.ts` → 3 tests fail, incl. "agent candidate can NEVER supersede a user entry" ✓
3. **Cap reservation** moved to post-await (naive check-then-increment) → the parallel-burst cap test fails ✓
4. **Executor owner-gate** disabled in `executors/prompt.ts` → the guest-reaction test fails ✓

All four tests genuinely guard their fix — none pass vacuously.

## Security chain (verified line-by-line)

- Every gate is bot-side and re-derived per call; the env flags in the MCP child only control tool *registration* and grant nothing. A forged/unknown `agent_action` degrades to `{ok:false}`, never a throw across the bridge.
- Unattended refusal fails **closed** across the upgrade boundary: pre-upgrade persisted routine/watch fires are caught by matching the prompt prefixes both runners genuinely stamp.
- The cap reservation is race-safe under JS interleaving (slot reserved synchronously before the awaited store write, refund only on duplicate/error).
- Owner gating on proposal decisions is two-layered: the executor blocks a non-owner **before** the pending prompt is consumed (a guest can neither approve nor veto), and lifecycle re-checks at save time.
- `singleLine()` collapses whitespace in model-authored card text; over-length input is refused rather than truncated. No forget/pause/delete/run actions exist in the union.
- Bridge sockets live in fresh 0700 `mkdtemp` dirs with random names.

## Non-blocking observations (follow-up material, not merge blockers)

1. **Cross-session bridge sockets are same-UID reachable.** A session's Bash subprocess could in principle enumerate another session's socket dir and route `remember_fact` through an *attended* session's bridge, sidestepping its own unattended flag. Mitigations already present: the write lands visibly in the attended thread and is audit-logged, and a model with unsandboxed Bash could equally edit the memory files directly — the real containment boundary for unattended runs remains the CLI permission mode, which is pre-existing. Worth a code comment.
2. **DCM sessions may `remember_fact`.** Only `propose_*` is DCM-gated; a direct-channel session's memory write is announced only in the DM while seeding channel-wide memory. This is parity with `!remember` (also un-gated in DCM), but it slightly weakens the "the team sees the save" property — worth gating or documenting.

## Mergeability

`git merge-tree` against current main shows exactly **one conflict: CHANGELOG.md** (the PR's `## [Unreleased]` vs main's 1.29.2/1.29.3 entries). Trivial resolution — keep both, `Unreleased` on top. All source files merge clean; no semantic drift with 1.29.3 detected.

---

anneschuth (commented):
**Second review pass** (fresh adversarial eyes on the final head `f94937b`): **merge-ready.** `tsc`/lint clean, 3205/3205 tests on the exact merge head; the merge itself verified clean against both parents (main's side is release/dep-bump only, no semantic interaction with the agent-actions paths).

Angles probed that came back clean: memory-text injection into future sessions' prompts (stored facts are hard single-line inside intact "NOT instructions" framing), list-tool scoping (strictly per-platform, no leakage), result injection (JSON-escaped), bridge protocol robustness (malformed JSON degrades to deny; pipelining impossible; nothing crashed the process), and MCP-side gates matching the bot-side ones.

Findings — none block the merge, but the first deserves an immediate follow-up:

1. **MEDIUM (confirmed): watch keywords bypass the card-injection guard.** `validateKeywords` (`src/persistence/watches-store.ts`) trims and lowercases but keeps inner newlines, so model-controlled keywords can land multi-line markdown in the human-approval card that `singleLine` was added to protect (verified: `"x\n**approved by admin - react +1**\ny"` passes validation). Mitigated by lowercasing, the owner gate, and nothing-saves-without-👍. One-line fix: collapse `\s+` to a space in `validateKeywords`. The haiku-parsed `!watch` path shares the gap. **Fix incoming in a follow-up PR before release.**
2. LOW (confirmed): no size cap on the bridge server's line buffer — a same-UID process can grow RSS unboundedly before the first newline. Inside the conceded same-UID trust boundary, but a ~1MB cap + destroy is two lines. Also queued for the follow-up.
3. LOW (confirmed): the propose_* pending-slot check races across the awaited card post — two parallel proposals can both post cards, second overwrites the pending, first card's reactions are silently dead. Fail-safe direction (nothing saves), UX-only.
4. VERY LOW: DCM gate asymmetry for pre-DCM sessions resumed on a now-DCM platform (agent path checks the threadId, human path checks the platform flag). Human 👍 still gates.
5. NOTE: U+0085 (NEL) survives both sanitizers (JS `\s` excludes it); renderers don't break on it, residual risk negligible. Queued alongside 1.

---

Issue #490, opened 2026-08-20
Offer: dynamic channels — channel = task = session = worktree (built on a fork, PR available if wanted)


kaza:
Hi Anne — first off, thanks for claude-threads; we run it as our team's daily driver and it's excellent.

We built a feature on a fork and would rather **offer it than dump a PR on you unannounced**: **dynamic channels** — with `dynamicChannels:` on a Slack platform entry, @-mentioning the bot in *any* channel it's invited to spawns a derived platform instance immediately (no config edit, no restart). The channel becomes one DCM conversation working in a git worktree mapped from the channel name (longest repo-prefix match) or a scratch dir. Archiving the channel tears it down mechanically: WIP-commit → push → verify clean-and-pushed → remove worktree (remote branches never deleted; failures keep everything and alert). A `workon` helper script lets the agent add worktrees for *additional* repos mid-task, recorded in a sidecar file the daemon reads at teardown.

Two design notes you may care about:

- Derived instances **never open a second Socket Mode connection** — your DM-discovery comment about Slack round-robining envelopes is exactly why. The parent keeps the only socket and injects events into derived clients; Web API calls are per-instance HTTP.
- Everything is opt-in: without `dynamicChannels` the behavior is unchanged; your full test suite passes (3,034 with our additions), plus new unit tests for the mapping, runtime lifecycle, and teardown.

Branch: https://github.com/vvs-capital/claude-threads/tree/dynamic-channels
Spec: https://github.com/vvs-capital/claude-threads/blob/dynamic-channels/docs/dynamic-channels-spec.md

If you'd like this upstream, we're happy to open a PR — as one piece or split (event routing / discovery runtime / teardown verifier), and to rework the shape to your taste. If it doesn't fit the project's direction, no hard feelings; we'll maintain the fork.

(One small standalone fix you might want regardless: sessions started *inside* an existing worktree currently get prompted by `worktreeMode: require` to create a worktree-in-a-worktree; our branch auto-detects and records instead.)

Jadefalkner:
Chiming in as the author of DCM (#478) and DM auto-discovery (#479).

kaza — the single-socket design (parent holds the only Socket Mode connection and injects events into derived instances) is exactly the right answer to the envelope-distribution problem that made me exclude Slack from DM discovery; nice to see the general case solved properly. The teardown principle (remote branch = durable store, worktree = rebuildable cache, remote branches never deleted) also reads very sound.

[...]

PR #501, opened 2026-08-25
fix: sessions started inside a worktree satisfy worktreeMode without prompting

kaza:
The first of the two PRs invited in #490 ("the worktree-in-worktree fix — please open that as its own small PR right away").

## Problem

A session whose `workingDir` already **is** a managed worktree (started there deliberately, or by an orchestration layer) still gets the worktree prompt under `worktreeMode: prompt`/`require` — offering to create a worktree inside a worktree. The isolation the mode exists for is already satisfied.

## Change

`shouldPromptForWorktree` now runs `detectWorktreeInfo(session.workingDir)` (after the existing early-outs, before the repo check — detection is a pure path-prefix check first, so no extra git spawns on the common path) and, when it detects one, records it on the session and skips the prompt.

Review hardening rounds (Gemini, CodeRabbit, Codex reviews before submitting) surfaced adjacent gaps this PR also closes:

- the detected worktree is **registered for reference counting** at the lifecycle call site, mirroring the resume path — otherwise another session's cleanup could `git worktree remove` it under a live session
- `messageManager.setWorktreeInfo` is synced like every other `worktreeInfo` assignment
- `detectWorktreeInfo` now resolves the **worktree root** (`rev-parse --show-toplevel`) instead of recording a nested working subdirectory as the worktree path, and treats a detached HEAD (`"HEAD"`) as no-branch
- detection runs even when the prompt is skipped (`skipWorktreePrompt` — e.g. unattended starts) so those sessions also get `worktreeInfo` and reference-count protection; the flag now suppresses only the prompt
- `cleanupSession` drops the worktree reference on **every** exit path (one-place rule, like account release) — pre-existing leak: an early Claude exit left a stale reference that blocked worktree cleanup until restart; this affected all registration paths, not just the new one

## Tests

New cases in `handler.test.ts`: detection records + skips prompt (require mode), messageManager sync, null-detection still prompts, off-mode runs no detection. Full suite green.

### DIFF
--- a/src/git/worktree.ts
+++ b/src/git/worktree.ts
@@ -335,8 +335,18 @@ export async function detectWorktreeInfo(
     // Get the branch name from git
     const branchOutput = await execGit(['rev-parse', '--abbrev-ref', 'HEAD'], workingDir);
     const branch = branchOutput?.trim();
-    if (!branch) {
+    // A detached HEAD reports the literal string "HEAD" — not a branch name;
+    // downstream (findWorktreeByBranch, list markers) would treat it as one.
+    if (!branch || branch === 'HEAD') {
       log.debug(`Could not detect branch for worktree at ${workingDir}`);
+      return null;
+    }
+
+    // The worktree root, not workingDir: a session may start in a nested
+    // subdirectory, and reference counting / `git worktree remove` must see
+    // the actual worktree path.
+    const toplevel = (await execGit(['rev-parse', '--show-toplevel'], workingDir))?.trim();
+    if (!toplevel || !isValidWorktreePath(toplevel)) {
       return null;
     }
 
@@ -347,9 +357,9 @@ export async function detectWorktreeInfo(
     log.debug(`Detected worktree: path=${workingDir}, branch=${branch}, repoRoot=${repoRoot}`);
 
     return {
-      worktreePath: workingDir,
+      worktreePath: toplevel,
       branch,
-      repoRoot: repoRoot || workingDir,
+      repoRoot: repoRoot || toplevel,
     };
   } catch (err) {
     log.debug(`Failed to detect worktree info for ${workingDir}: ${err}`);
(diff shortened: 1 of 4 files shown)


anneschuth:
Thanks so much! I'll only be able to look into this on Friday, I'm travelling and my deploy key expired, I'll need to sit behind a computer to fix it. Sorry for the wait!

anneschuth (approved):
Thank you for this — it does exactly what #490 invited, and the execution is careful. Reviewed with independent verification in a scratch worktree against current main; **approving**.

## Verified

- **Detection correctness**: `isValidWorktreePath` runs before any git spawn, so a normal repo pays zero extra processes; the `startsWith(WORKTREES_DIR + path.sep)` check can't false-positive on sibling dirs. Detached-HEAD handling (`rev-parse --abbrev-ref HEAD` → literal `HEAD` treated as no-branch) is right, and re-validating `--show-toplevel` also fixes the pre-existing resume path that could store a nested `workingDir` as `worktreePath`.
- **The refcount leak you fixed in `cleanupSession` is real** — verified on main: a session that registered a worktree and whose Claude died early (the "Exited before Claude responded" path) left a stale `worktreeUsers` entry until bot restart, blocking `!worktree cleanup`. Nice catch beyond the invited scope.
- **Release safety**: `unregisterWorktreeUser` is an idempotent Set delete, `cleanupSession` is terminal, and the pause/`!cd`-restart/kill-without-unpersist paths don't go through it — no release-under-a-live-session hazard.
- **Tests**: `tsc` clean, full unit suite 3155/3155 green with the PR. Red-check passed: with main's handler and your tests, "records the worktree and skips the prompt when the session starts inside one" genuinely fails — the main new test guards the fix.
- **Mergeability**: merges cleanly onto current main (`git merge-tree` clean).

## Non-blocking notes (fine as follow-ups)

1. The lifecycle changes (the `cleanupSession` unregister and the start-time registration) don't have their own regression tests yet — `mock-session-context.ts` already exposes `unregisterWorktreeUser`, so a small test would be cheap if you feel like it, but I won't hold the merge on it.
2. Behavior nuance we're accepting: on resume-failure with retries left, the reference is now released, so a co-using session could clean up that worktree before the failed session's retry after restart. Marginal and arguably more correct.
3. Perf nit: on `skipWorktreePrompt` paths, detection now runs up to 3 git spawns where main ran zero. Fine.

Will merge once the two in-flight integration checks come back green.

---

PR #502, opened 2026-08-25
Slack: shared event source — one Socket Mode connection per app

kaza:
The second PR invited in #490: the single-socket event-injection layer, extracted standalone — routing + tests, none of the dynamic-channels semantics.

## Problem

Slack round-robins Socket Mode envelopes across **all** of an app's open connections. Any second `SlackClient` on the same app token therefore silently steals events from the first — the reason #479's DM auto-discovery excludes Slack today.

## Design

Exactly one client (the parent) owns the socket; other clients become **secondaries**:

- a secondary is constructed with the parent as an optional constructor argument; its `connect()` registers with the parent and opens no socket (Web API calls, which are plain HTTPS, work independently)
- the parent's `handleSlackEvent` hands events for registered channels (`event.channel` or `event.item.channel`) to the owning secondary via `_injectSlackEvent` **before** its own-channel filtering; the injected event then flows through the secondary's normal pipeline (bot-message filter, dedup, emission)
- events for unregistered channels keep today's behavior byte-for-byte; a config without secondaries has an empty routing map — **zero behavior change for existing deployments**

Lifecycle, from the pre-submission review rounds (Gemini, CodeRabbit, Codex):

- the parent mirrors `connected` / `disconnected` / `reconnecting` onto secondaries — its socket state IS their state
- on parent reconnect, missed-message recovery runs for the parent **and** every registered secondary
- `disconnect()` unregisters with an identity check, so an old instance's teardown can't evict its replacement; a disconnect racing `connect()`'s `fetchBotUser` await can't resurrect a dead registration
- registering a secondary for the parent's own channel throws (it could never receive events)
- documented ordering: register secondaries before connecting the parent, so no events fall in the gap

## Deliberately not included

No production consumer is wired — per the scoping in #490 this is the mechanism; DM auto-discovery on Slack (or any multi-channel feature) would be the consumer.

## Pre-existing caveat worth knowing (not introduced here)

`getMcpConfig()` hands the app token to the MCP permission server, which opens its own Socket Mode connection during permission prompts (`mcp-platform-api.ts`) — while that socket is open, Slack round-robins events to it too. Making the one-socket invariant absolute would mean routing those through the parent as well; happy to look at that as a follow-up.

## Tests

`shared-event-source.test.ts`, 10 cases: no-socket connect, routing (messages + reaction items), own-channel unaffected, unregistered-channel drop path unchanged, unregister stops routing, state mirroring, replacement-eviction guard, same-channel rejection, end-to-end emission with dedup on the injected path, disconnect unregisters. Full suite green (3,454).

### DIFF
--- a/src/platform/slack/client.ts
+++ b/src/platform/slack/client.ts
@@ -117,16 +117,26 @@ export class SlackClient extends BasePlatformClient {
    * A secondary's only event feed is the parent's socket, so the parent's
    * connection state IS the secondary's connection state — mirror it.
    *
-   * Re-armed after `disconnect()`, whose base teardown detaches every
-   * listener including this one.
+   * Stable handler identities, so re-arming is idempotent: overlapping
+   * `disconnect()` calls each clear the listeners and each re-arm, and a
+   * fresh closure per install would leave the mirror stacked and every
+   * later event duplicated into the secondaries.
    */
-  private installStateMirror(): void {
-    for (const state of ['connected', 'disconnected', 'reconnecting'] as const) {
-      this.on(state, (...args: unknown[]) => {
+  private readonly stateMirrors = (['connected', 'disconnected', 'reconnecting'] as const).map(
+    (state) => ({
+      state,
+      handler: (...args: unknown[]) => {
         for (const secondary of this.channelClients.values()) {
           secondary.emit(state, ...args);
         }
-      });
+      },
+    })
+  );
+
+  private installStateMirror(): void {
+    for (const { state, handler } of this.stateMirrors) {
+      this.off(state, handler);
+      this.on(state, handler);
     }
   }
 
@@ -180,20 +190,20 @@ export class SlackClient extends BasePlatformClient {
     this.handleSlackEvent(event);
   }
 
-  override async disconnect(): Promise<void> {
+  override disconnect(): Promise<void> {
     // A secondary must stop receiving injected events when it goes away;
     // the base teardown is still safe to run (it has no socket to close).
     if (this.sharedEventSource) {
       this.sharedEventSource.unregisterChannelClient(this.channelId, this);
     }
-    try {
-      await super.disconnect();
-    } finally {
-      // Base teardown calls removeAllListeners(), which takes the state mirror
-      // with it. Without re-arming, a parent that goes intentional-disconnect
-      // → reconnect never tells its secondaries about a connection again.
-      this.installStateMirror();
-    }
+    // Base teardown calls removeAllListeners() synchronously and then returns
+    // the socket-close promise, which can stay pending. Re-arm here rather
+    // than after awaiting it: without the mirror, a parent that reconnects
+    // while the close is still settling emits 'connected' into nothing and
+    // its secondaries miss it permanently.
+    const closed = super.disconnect();
+    this.installStateMirror();
+    return closed;
   }
 
   // ============================================================================
--- a/src/platform/slack/client.ts
+++ b/src/platform/slack/client.ts
@@ -197,13 +197,20 @@ export class SlackClient extends BasePlatformClient {
       this.sharedEventSource.unregisterChannelClient(this.channelId, this);
     }
     // Base teardown calls removeAllListeners() synchronously and then returns
-    // the socket-close promise, which can stay pending. Re-arm here rather
-    // than after awaiting it: without the mirror, a parent that reconnects
-    // while the close is still settling emits 'connected' into nothing and
-    // its secondaries miss it permanently.
-    const closed = super.disconnect();
-    this.installStateMirror();
-    return closed;
+    // the socket-close promise, which can stay pending. Re-arm before that
+    // promise settles: without the mirror, a parent that reconnects while the
+    // close is still in flight emits 'connected' into nothing and its
+    // secondaries miss it permanently.
+    //
+    // Synchronous try/finally, not async: the base promise is returned
+    // unchanged (a synchronous throw stays synchronous) and the mirror is
+    // re-armed on the throwing path too — a failed close must not also leave
+    // the parent permanently mute.
+    try {
+      return super.disconnect();
+    } finally {
+      this.installStateMirror();
+    }
   }
 
   // ============================================================================
--- a/src/platform/slack/client.ts
+++ b/src/platform/slack/client.ts
@@ -83,8 +83,26 @@ export class SlackClient extends BasePlatformClient {
 
   private readonly formatter = new SlackFormatter();
 
-  constructor(platformConfig: SlackPlatformConfig) {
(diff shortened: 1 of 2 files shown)


anneschuth (changes requested):
Thank you — the architecture here is right, the scope is exemplary (+303/−6 across exactly 2 files, no production consumer wired yet, zero behavior change for existing configs — all independently verified), and the race guards are real. Two issues block the merge, though; both should be quick to fix.

## Blocking

**1. Secondary missed-message recovery never runs — the block is dead code.**
In `src/platform/slack/client.ts` (hello handler, ~line 426–442): the handler calls `this.onConnectionEstablished()` first and *then* checks `if (this.isReconnecting)`. But `BasePlatformClient.onConnectionEstablished()` already runs the parent's recovery and **unconditionally sets `isReconnecting = false`** before returning — so the entire block after it, including the new per-secondary `recoverMissedMessages()` loop, is unreachable. (The parent's inner branch there was already dead on main, which masked this — parent recovery works via the base method; the PR extended a pre-existing dead block.) Net effect: after a reconnect, secondaries silently lose their missed messages — exactly the scenario the comments promise to handle — and no test covers it.

Fix: run the secondary recovery loop **before** calling `onConnectionEstablished()`, or override `onConnectionEstablished()` in `SlackClient`. Please add a red-capable test for it (one that fails if the loop is removed).

**2. The PR's own test suite fails (9/10) — and CI's `test` job is red because of it.**
`shared-event-source.test.ts:66` expects `connected === true` after `secondary.connect()`, but the code deliberately withholds `'connected'` unless the parent's socket is live (the `socketConnected` gate). The code's behavior looks like the intended one, so the test should assert `false` there — or simulate the parent's hello first if you want to exercise the deferred emission.

## Non-blocking, but recommended while you're in there

3. The dedup assertion at `shared-event-source.test.ts:181-184` isn't red-capable: I removed the dedup guard and it still passed, because the duplicate's `'message'` emission is deferred through `getUser().then(...)` and the synchronous `toHaveLength(0)` runs before the microtask fires. An async flush before asserting fixes it (the first half of that test already awaits properly).
4. Latent interaction worth a note or fix before a production consumer lands: the state-mirror listeners are installed once in the constructor, but base `disconnect()` calls `removeAllListeners()` — a parent that goes intentional-disconnect → reconnect loses the mirror permanently. Latent today (no consumer, and auto-reconnect doesn't remove listeners).

## What was verified independently

Scratch worktree on your head `d0b5699`: typecheck clean; full unit suite 3161 pass with only your new test failing (item 2). Removing the routing block turns 4 of your tests red — those are genuine regression guards. Zero-behavior-change claim confirmed (routing gate falls through on the `Map.get` miss; foreign channels hit the identical drop paths; parent dedup/`lastProcessedTs` untouched). Merges cleanly onto current main.

Looking forward to the fixes — with 1 and 2 addressed this is a merge.

---

anneschuth (approved):
Thank you — both blockers are resolved exactly right, and the extras go beyond what was asked. Re-verified independently in a scratch worktree; **approving**.

- **Secondary recovery is now real code**: the `onConnectionEstablished` override captures `wasReconnecting` *before* `super` clears it, then recovers every registered secondary — and the old dead block in the hello handler is gone with a comment explaining why. Red-verified: reverting the capture makes "recovers missed messages for every registered secondary" genuinely fail; restored, 17/17 pass. The first-connect negative case is covered too.
- **The test suite is green as shipped**: full unit run 3169/3169 in the worktree, `tsc`/`eslint` clean, and all of this PR's CI (integration on both platforms included) green on this head.
- **The mirror/`removeAllListeners()` latent issue is fixed properly**, not just documented: stable handler identities make the re-arm idempotent, and the synchronous `try/finally` around `super.disconnect()` re-arms even on the throwing path — the two follow-up commits show the care taken here.
- **Merges clean onto current main** (1.31.2).

The mechanism is in — DM auto-discovery on Slack (or any multi-channel consumer) now has its single-socket foundation. Merging.

---

kaza:
Thanks — this was a genuinely useful review, and the two blocking items were both real.

## Blocking

**1. Dead secondary recovery.** Fixed by overriding `onConnectionEstablished()` in `SlackClient`: it captures `isReconnecting`, delegates to `super` (which recovers this client's own messages and clears the flag), then recovers each registered secondary.

I also deleted the block in the hello handler rather than reordering it. As you noted, the parent's own branch there was already dead on main — the base recovers this channel unconditionally, and `recoverMissedMessages()` self-guards on `lastProcessedTs`, so the outer `if (this.lastProcessedTs)` was adding no semantic guard either. The override is now the single place reconnect recovery happens.

Two red-capable tests: `recovers missed messages for every registered secondary` fails if the override is removed, and `does not recover for secondaries on a first connect` pins the negative branch.

**2. Failing test.** You were right that the code's behaviour was the intended one. The test now asserts `false` immediately after `connect()`, then asserts the deferred emission arrives with the parent's hello — plus a new case for the fast path where the parent's socket is already live.

## Non-blocking, both taken

**3.** The dedup assertion now flushes the microtask queue before asserting, so removing the guard makes it fail.

**4.** Fixed rather than noted. `installStateMirror()` is extracted and re-armed after teardown.

## What our own review then found in my fixes

Worth reporting, because all three were introduced by the fix in (4), not by the original PR:

- **Overlapping `disconnect()` stacked the mirror.** Fresh closures per install meant two concurrent disconnects each cleared and each re-installed — six listeners, every later event duplicated into the secondaries. Reproduced with `Promise.all([parent.disconnect(), parent.disconnect()])`.
- **The mirror was absent for the whole async close window.** My first attempt re-armed in a `finally` after awaiting the socket close, which can stay pending for the safety timeout. A reconnect landing in that window emitted `connected` into a mirror-less parent and the secondaries missed it permanently.
- **A synchronous throw from `super.disconnect()` skipped the re-arm** entirely, leaving the parent permanently mute towards its secondaries.

Final shape: stable handler identities with an idempotent `off()`-then-`on()` install, and a **synchronous** `try/finally` rather than `async`/`await` — so `disconnect()` still returns the base promise unchanged, a synchronous throw stays synchronous, and the re-arm covers every path including the failing one. Each of the three has a regression test that is red against the commit before it.

## Two things I did not change

- **Guarding `installStateMirror()` on secondaries.** It does run there, and `channelClients` is always empty, so it attaches three no-op listeners per secondary. I left it: the guard trades those for a role-conditional in the constructor and silently encodes "secondaries never have secondaries", an invariant nothing else enforces. Happy to add it if you'd rather have it explicit.
- **`does not recover for secondaries on a first connect` is not red-capable** against the override — deliberately. It guards the negative branch so that a future "just always recover" change fails.

Suite is 3251 pass / 0 fail and `tsc --noEmit` is clean locally, so CI's `test` job should go green with this push.

PR #512, opened 2026-08-29
Security review fixes: watch/routine approval posture + hardening

anneschuth:
## Summary

A full-codebase security review turned up one higher-risk area (the watch/routine → unattended-session trust model) and a handful of lower-severity hardening items. This PR fixes all of them. Every fix carries a regression test (verified red-green where applicable); the full suite, typecheck, lint and knip are green.

The core review conclusion was reassuring: no exploitable path traversal, SSRF, auth bypass, prototype pollution or ReDoS. The `send_file` path validator, permalink scoping, decision bridge, the two authorization primitives, and YAML/JSON deserialization are all sound. The changes below close the concrete issues that remained.

## Changes

**Watch/routine trust model**
- **Explicit approval posture, chosen at creation.** The confirmation card now offers 👍 *save* (each fired run asks for in-thread approval before every tool action) or ✅ *save + run autonomously* (no approval prompts), alongside 👎 *discard*. The choice is persisted per item (`requireApproval`) and enforced at fire time via `forceInteractivePermissions`, so an approval-required fire runs interactive **even on a `skipPermissions` platform** — a watch triggered by attacker-influenceable channel content can no longer silently execute tools with no human in the loop. Safe posture is the default; existing/older items and agent-proposed items always require approval (the autonomous option is never offered for agent proposals).
- **Distillation skips unattended (routine/watch-fired) sessions**, matching the existing `remember_fact` guard — a prompt-injected fire can no longer persist attacker-derived "facts" from its thread into channel memory (which is injected into every future session's system prompt).
- **The author identity in watch confirm/fire prompts is collapsed to one line** so a future platform's free-form display name can't smuggle newlines/fake delimiters out of the quoted block.

**Access control**
- **The `✅ Invite to session` reaction is now owner-gated**, closing an asymmetry with the owner-gated `!invite` command: a temporarily-`!invite`d guest could previously grant *standing* membership to an unauthorized third party by reacting on their message-approval card. A non-owner's ✅ is now downgraded to a one-shot allow.

**Injection / platform**
- **Worktree branch names are validated at the `createAndSwitchToWorktree` chokepoint** (the in-session `!worktree <name>` path skipped it). `isValidBranchName` now also rejects shell metacharacters (`& | ; $ \` ( ) < > ! ' " # %`) — git permits these in ref names but they become a command-injection vector on Windows (spawn wrapper uses `shell:true`); `git worktree add` gained a `--` separator against flag injection.
- **`findByThreadIdAnyState` is scoped by `platformId`** on the plain-reply resume path, so a thread-id collision across platforms can't resume another platform's session (its allowlist, working dir, worktree, Claude account).

**Persistence hardening**
- **`update-state.json` is written owner-only (0600)** via the shared atomic writer (previously 0644).
- **Persisted free-text fields (`firstPrompt`, `queuedPrompt`, `lastTasksContent`) are capped** before entering `sessions.json` (rewritten on every mutation); the cap is far above any real prompt.

Intentional design decisions flagged by the review — Mattermost `read_post` returning open-channel posts, `send_file` bypassing per-call approval, empty-allowlist = everyone, and invited guests being able to `!stop` — were left as-is.

## Testing

- Full unit suite: **3268 pass, 0 fail** (`bun run test`).
- `bun run typecheck`, `bun run lint`, `bun run knip` all clean.
- Each fix has a regression test; the highest-value ones (invite-downgrade, distillation guard, worktree chokepoint) were verified **red** without the fix and **green** with it.
- Confirmed empirically that `git worktree add -- <path> <branch>` and `-b <branch> -- <path>` are accepted.

## Checklist

- [x] Documentation updated (if needed) — CHANGELOG `[Unreleased]` + CLAUDE.md Routines/Watches sections

---
_Generated by [Claude Code](https://claude.ai/code/session_01K9MLgT2BbGTdtM6DuMgD9a)_


anneschuth (commented):
Deep review with independent verification — **merge-ready**. (Comment review; the token can't approve its owner's PR.)

## Verified on head `4f3b37b`

- `tsc`, `eslint`, `knip` clean; **3270/3270 tests** in a clean worktree.
- **All eight claimed changes confirmed against the actual diff**, including the load-bearing plumbing: `forceApproval: item.requireApproval ?? true` at both fire sites → `forceInteractivePermissions` → `permissionMode='default'`, verified to override `skipPermissions` platforms; the manual `!routines run` path rides the same chain.
- **Backward compat is solid at four independent layers**: store read normalization, store `add()`, the save path, and both fire sites all default a missing `requireApproval` to approval-required. Old persisted approvals without `sessionOwner` fall back to platform-allowlist only — fail-safe. A resumed approval-required fire keeps interactive mode after bot restart.
- **Respawn plumbing**: the sticky `forceInteractivePermissions` flag is carried by the pre-existing `sessionHasInteractiveOverride` on `!cd`/`!permissions`/worktree respawns — no new required field needed, invariant satisfied by construction.
- **Emoji distinctness**: `+1`/`thumbsup` vs `white_check_mark`/`heavy_check_mark` are disjoint after normalization on both platforms, approval is checked first, and misordering could only degrade toward the safe posture.
- **Red-green spot checks all genuinely red**: invite-downgrade removed → 1 fail; distillation unattended-guard removed → 1 fail (attended test still green, proving discrimination); branch-name chokepoint neutralized → exactly the new injection test fails. All restored.
- **Windows injection vector is real and closed**: `spawn.ts` uses `shell:true` on win32 and `execGit` rides that wrapper; legitimate names (`feature/login`, dots, `@`) still pass.
- Scope: every one of the 37 files maps to the described changes + tests + docs.

## Non-blocking follow-ups worth an issue or a small PR

1. **Active-session threadId lookups remain cross-platform-unscoped** (`SessionRegistry.findByThreadId`, `isUserAllowedInSession`'s active check) — same defense-in-depth class this PR fixed for the persisted store; real IDs don't collide today.
2. **Posture is invisible and immutable after creation**: `!routines`/`!watches` listings don't show `requireApproval` and there's no flip command — an owner auditing existing watches can't see which are autonomous. A posture marker in the list output would be cheap.
3. Cosmetic: a non-owner's ✅ downgrade is silent to the reactor (they may believe they invited the user), and a rejected branch name with a backtick can escape its code span in the error post. Display-only.

Merging per the standing order.

---

PR #514, opened 2026-08-29
1.31.1: regression guard for cross-platform resume scoping + two display fixes

anneschuth:
## Summary

Follow-ups to the **1.31.0** security release. The final focused review of the review-fix delta landed *after* 1.31.0 had already auto-released; this patch addresses its one finding plus two cosmetic nits the owner's merge-review flagged. Nothing here fixes a bug in 1.31.0 — the shipped code is correct — but it adds a missing regression guard and two display fixes.

**This PR bumps the version (`1.31.0 → 1.31.1`), so merging it will trigger `release.yml` → npm publish of 1.31.1.**

## Changes

- **Regression guard for the cross-platform resume scoping** — `resumePausedSession` → `findPersistedByThreadId` is now covered by a red-green test: a message on one platform must not resume a session persisted under another platform whose `threadId` collides. Verified it fails when the `session.platformId === platformId` check is removed (the 1.31.0 fix shipped correct but untested; the review flagged this as the one gap).
- **Sanitize a rejected branch name in the error post** — an invalid `!worktree <name>` containing a backtick/newline is stripped for display so it can't break out of its markdown code span.
- **Non-owner `✅ Invite` downgrade is no longer silent** — when a guest's ✅ is downgraded to a one-shot allow (only the owner may grant standing membership), the bot now says so instead of leaving the reactor to assume the invite worked.

Patch level per semver: fixes + a test, no new feature.

## Testing

- Full unit suite: **3271 pass, 0 fail** (`bun run test`).
- `bun run typecheck`, `bun run lint`, `bun run knip` all clean.
- The new scoping test verified **red** (revert the scope check → fails) and **green** (restored → passes).

## Not in this PR (tracked as follow-ups)

- Active-session `threadId` lookups (`findByThreadId`, the active branch of `isUserAllowedInSession`) remain cross-platform-unscoped — same defense-in-depth class, IDs don't collide today.
- The approval posture isn't shown in `!routines`/`!watches` listings and can't be flipped after creation.

## Checklist

- [x] Documentation updated (if needed) — CHANGELOG `[1.31.1]`

---
_Generated by [Claude Code](https://claude.ai/code/session_01K9MLgT2BbGTdtM6DuMgD9a)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.31.1] - 2026-08-29
+
+### Security
+- **Regression guard for the cross-platform resume scoping (1.31.0).** The `resumePausedSession` sink is now covered by a test that fails if the `platformId` scope is removed — a message on one platform must never resume a session persisted under another platform whose thread id collides. The fix shipped correct in 1.31.0 but without this guard.
+
+### Fixed
+- **A rejected branch name can no longer break its own error message.** An invalid `!worktree <name>` whose name contains a backtick or newline is sanitized for display, so it stays inside its markdown code span in the error post.
+- **A downgraded "✅ Invite to session" reaction is no longer silent.** When a non-owner participant's ✅ is downgraded to a one-shot allow (only the owner may grant standing membership), the bot now says so, instead of leaving the reactor to assume the invite succeeded.
+
 ## [1.31.0] - 2026-08-29
 
 ### Added
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "claude-threads",
-  "version": "1.31.0",
+  "version": "1.31.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "claude-threads",
-      "version": "1.31.0",
+      "version": "1.31.1",
       "license": "Apache-2.0",
       "dependencies": {
         "@hono/node-server": "2.1.1",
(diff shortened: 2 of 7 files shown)


anneschuth:
CI: `integration (mattermost)` failed on `c69c362`, on one test:

> `Session Lifecycle > mattermost platform > Follow-up Messages > should ignore side conversations (@other_user)` — `expect(afterBotPostCount).toBe(initialBotPostCount)` (an extra bot post appeared).

**This is not this PR's failure — it's a pre-existing timing flake.** This PR's only production changes are the invite-downgrade notice (`message-approval.ts`) and the invalid-branch-name display sanitizer (`worktree/handler.ts`). The failing test starts a session, ends it, and sends a plain non-mention follow-up — it exercises neither code path. The only conceivably related change, the `resumePausedSession` `platformId` scoping, already landed on `main` in #512 and is a no-op when the platform matches (as it does in a single-platform integration run), so it can't change this test's behavior. The test file itself is unmodified here (untouched since #464). The run also logged `Failed to add reaction 👍/👎: 404 couldn't find the emoji`, i.e. environment noise in this run.

I attempted the one allowed re-run of the failed job but the integration lacks permission (`403 Resource not accessible by integration`), so I can't confirm-by-re-run myself. I'm not going to modify an unrelated integration test to force green. A maintainer re-run should clear it. Keeping the PR watched until it's green/merged.

---

anneschuth (commented):
Reviewed — this is precisely the three follow-ups from the #512 review, each implemented the way I'd have asked: the cross-platform resume scoping now has a discriminating regression test at the sink (the one gap the review flagged — the fix shipped correct but unguarded), `displayBranchName` strips exactly the two break-out characters for display while rejection still happens on the raw value, and the ✅ downgrade now tells the reactor what actually happened.

#515 (the mattermost test de-flake that was the only red here) is merged; I've updated this branch to pick it up and will merge once CI is green — which auto-releases **1.31.1**.

---

Issue #505, opened 2026-08-25
Feature: per-platform option to quiet tool-call streaming and lifecycle status posts (assistant-style use)


thejdubb02:
I use claude-threads on Mattermost as a personal assistant. One long-lived direct-channel-mode channel I talk to through the day, not for reviewing coding sessions. It's great for that, but almost all of the channel output is designed for watching Claude *work*, and there's no way to turn it down. Two sources dominate:

**Tool-call streaming** (`src/operations/transformer.ts`). Every tool use renders into the channel (`Bash ...`, `Read ...`, `Edit ...`), each with a `↳ ✓` completion indicator. In a coding session that's the whole point. As a chat assistant it's the bulk of what shows up, and the actual reply is buried under the commands that produced it. The existing `sessionHeader` / `stickyMessage` verbosity fields don't touch it.

**Lifecycle status posts** (`src/session/lifecycle.ts`). `Session idle - will timeout in ~N minutes`, `Session timed out after N minutes`, and `Session resumed ...` each post to the channel, and on mobile each is a push notification. For a bot that's just meant to be there when I message it, the session lifecycle is an implementation detail I don't need narrated. (#491 argues these status posts shouldn't wake *other bots*; this is the user-facing side of the same thing.)

**Proposal:** a per-platform option in the spirit of the existing `stickyMessage` / `sessionHeader` fields:

```yaml
- name: assistant
  type: mattermost
  toolActivity: hidden       # full (default) | hidden
  lifecycleNotices: hidden   # full (default) | hidden
```

`toolActivity: hidden` suppresses the streamed tool blocks and their completion indicators (Claude's text replies still post). `lifecycleNotices: hidden` suppresses the idle/timeout/resume posts. The session still times out and resumes exactly as now, it just isn't announced. Errors keep posting regardless.

I've been running this locally against 1.29.0 by patching the built bundle (gating the tool-display ops in the transformer, and short-circuiting the `timeout`/`resume` post kinds), and it does exactly what I want: replies only. I patched `dist` rather than `src`, so I'd rather match your config conventions than guess. Happy to open a PR if the shape looks right.

Thanks for building this. It makes a genuinely good always-on assistant.

anneschuth:
Thank you — the assistant-style use case is real, and you're right that today all the output is designed for watching Claude *work*. Your proposed shape fits the config conventions well, so: yes, a PR would be very welcome.

Two notes to save you a round-trip:

[...]

PR #525, opened 2026-09-02
Slack: show a working status instead of a no-op typing indicator

kaza:
## Summary

`sendTyping()` is a no-op on Slack. Its comment — *"Slack doesn't have a typing indicator API for bots"* — was true when it was written, and is why Mattermost sessions show activity while Slack sessions show nothing between the ack reaction and the first output.

Slack does now have an equivalent: `assistant.threads.setStatus` renders a live `<App> is …` line under the app name, rotates `loading_messages` while it's up, and needs only `chat:write`. No manifest change, no new scope, nothing for existing installs to do.

This fills in the no-op, and adds the missing other half of the contract.

⚠️ **One thing you should weigh before merging** — see *The documentation disagrees* below. I'd rather flag it than have you find it.

## Changes

- **`clearTyping()` on `PlatformClient`**, defaulting to a no-op in `BasePlatformClient`. `stopTyping()` now calls it, and only when a timer was actually running.

  This isn't incidental. The existing contract assumed a *self-expiring* indicator — a Mattermost websocket frame stops on its own, so clearing the local interval was enough. A Slack status persists until something replaces it, so without this a finished session claims to be working forever. I shipped the first version without it and my own users found it within the hour. Platforms that don't need it inherit the base no-op and are unaffected.

- **`sendTyping()` implemented for Slack**, fire-and-forget: a status is a nicety and must never interfere with the work it describes. Workspaces where the method is unavailable log once at debug and carry on.

- **`src/platform/slack/status.ts`** — two pure helpers, so the parts worth testing don't need a socket:

  - **`statusAnchor()`.** `setStatus` requires a real message ts. A classic thread anchors to itself, but in direct-channel mode the session's thread id is synthetic and `resolvePostThreadId` already refuses to send it to Slack — and there's no thread to anchor to at all. It falls back to `lastProcessedTs`, the message that triggered the work. Without this the feature simply doesn't exist for DCM users.

  - **`dueForRefresh()`.** `startTyping` ticks every 3s, which is right for a websocket frame and wrong for an HTTP Tier-3 method — a few concurrent sessions would spend the rate limit re-asserting an unchanged status. It's a 20s heartbeat instead, with the anchor map pruned so it can't grow with every thread the bot has ever seen.

## The documentation disagrees

Slack documents `assistant.threads.setStatus` as working in **AI-assistant threads only**, and even names the error you should get elsewhere: `method_not_supported_for_channel_type`.

That is not what happens. Called against an ordinary public channel with a bot token carrying only the scopes this project already requests, it returns `ok: true` — on a threaded message and on a top-level one — and renders. I verified this against live Slack before writing any code, and it's been running on our deployment since.

So this relies on behaviour that contradicts Slack's own docs. Either the docs lag the implementation, or the behaviour could be tightened later. The failure mode if Slack does tighten it is mild — `setStatus` starts erroring, the `.catch` logs at debug, and Slack sessions go back to exactly today's behaviour — but you should decide whether you want that dependency in the codebase, and I'd understand either answer.

`agents.sessions.setStatus` also exists on the API and is **not in the public docs at all** (404). Its `status` is an enum, not free text; fifteen plausible values were all rejected. I mention it only so nobody else spends an evening on it.

## Why this might be worth having

#505 asks for a way to quiet tool-call streaming for assistant-style use. That reporter is on Mattermost, where the typing indicator already provides a liveness signal — so quieting the stream there still leaves *something*.

On Slack it doesn't: quiet the stream today and the channel goes completely silent while the bot works. This gives Slack the signal Mattermost already had, which makes that option safe to offer on both platforms rather than one.

## Testing

- `bun run test` — 3385 pass / 0 fail
- `bun run typecheck` and `bun run lint` clean
- New unit tests: anchor selection for both thread shapes and the no-anchor case; the throttle's first-send, tick-suppression and lapse behaviour; and `stopTyping` telling the platform to clear (plus not doing so when it never started).
- Live: verified `setStatus` and the empty-status clear against real Slack, and running on our deployment across ordinary channels and direct-channel-mode channels.

## Checklist

- [ ] Documentation updated (if needed) — happy to add a line to `docs/CONFIGURATION.md` or the platform implementation guide if you want the `clearTyping` contract written down; I didn't want to presume the shape.


anneschuth (changes requested):
Thank you — this is another genuinely well-built PR, and the write-up (especially flagging the docs contradiction yourself rather than letting us find it) is exactly how a dependency like this should be proposed. We verified it independently on a scratch worktree of 32a9cff against current main:

**What we verified:**
- `tsc`, lint clean; unit tests **3303 pass / 0 fail**; Slack integration suite 136 pass / 0 fail.
- **Red-checks 3/3**: reverting the `clearTyping` call in `stopTyping`, making the clear unconditional, and gutting `statusAnchor` each made the corresponding new tests fail — these are real regression tests.
- **Fire-and-forget confirmed**: both `api()` calls get `.catch` attached synchronously; `clearTyping` is never awaited at any of the ~12 `stopTyping` call sites, so teardown can't block. Base no-op means Mattermost is untouched.
- **Throttle and prune confirmed sound**: throttle set before the call (a failing workspace retries only every 20s), map bounded at 64 + recently-active anchors, `clearTyping` deletes its own key. Per-instance state means no interference with the shared event source from #502.
- Fun fact: our integration mock server has no `assistant.threads.setStatus` handler and 404s every call — so the whole suite was accidentally a live test of the degradation path, and nothing broke.
- Rebased directly on current main, scope exactly as described.

**The one change we'd like before merging:**

The PR promises debug-only degradation, but `api()` itself warn-logs any `ok:false` whose code isn't in `expectedErrors` — and these two calls pass none. If Slack ever tightens the method (`method_not_supported_for_channel_type`, which their docs explicitly name) or a workspace lacks it, every active session would emit a warn roughly every 20 seconds plus one per clear, indefinitely. Could you pass `expectedErrors: ['method_not_supported_for_channel_type', 'missing_scope', 'invalid_thread_ts', 'channel_not_found']` on both `assistant.threads.setStatus` calls (the set in `sendTyping` and the clear in `clearTyping`)? That makes the degradation genuinely silent, as the PR's own framing intends.

**Non-blocking notes:**
- In direct-channel mode the anchor is `lastProcessedTs`, which advances on every non-bot channel message — so a heartbeat can assert the status on anchor A, a later one on anchor B, and the clear only clears B; A's status lingers in that message's thread until Slack's auto-clear-on-reply catches it. Low visibility and self-limiting, so a short code comment (or clearing the previous anchor when it changes) is plenty — fine as a follow-up.
- The PR body's test count (3385) looks stale — we count 3303 (main's 3295 + the 8 new tests). Immaterial, just noting it so nobody chases the discrepancy.
- On the undocumented-API dependency itself: we're comfortable taking it. The degradation path is now empirically proven (see the mock-server accident above), and with `expectedErrors` in place a Slack-side withdrawal costs nothing but the feature itself.

- The offer to document the `clearTyping` contract in the platform implementation guide is a good one — a couple of lines there would be welcome, in this PR or a follow-up, whichever you prefer.

Once the `expectedErrors` change is in, this is good to merge, and we'd like to ship it in the next release together with #527.

---

anneschuth:
Status, so this doesn't look stalled: **one change left, and it's the `expectedErrors` array.**

Pass `['method_not_supported_for_channel_type', 'missing_scope', 'invalid_thread_ts', 'channel_not_found']` on both `assistant.threads.setStatus` calls (the set and the clear). Nothing else from my review is blocking — the anchor-lingering note and the implementation-guide docs are both fine as follow-ups.

I want this in the next release. If you'd rather I pushed the one-liner to your branch myself, say so and I will.

anneschuth:
Took you up on my own offer and pushed the `expectedErrors` change rather than making you round-trip for an array — you've got seven threads open here and this was the only thing left on this one.

Both `assistant.threads.setStatus` calls now pass the four documented codes, as a shared constant with the reasoning next to it. Tests cover each code plus a control asserting an unexpected error still warns; red-checked (emptying the list fails all four, control stays green). Observed through `setLogHandler` rather than the console, since the logger binds that at module load.

3308 tests pass, tsc and eslint clean.

Nothing else outstanding from my review — the anchor-lingering note and the implementation-guide docs are both fine as follow-ups. Shout if you'd rather I hadn't pushed to your branch; I'll keep to comments in future if you prefer.

anneschuth:
Also merged current main in — #549 landed in between and collided with this branch. It was a pure append collision in `client.test.ts` (your new bot-authored-filter block and my new status block both appended at the end); both are kept, nothing was dropped. 3390 tests pass on the merge.

All 10 checks green.

anneschuth (approved):
Approving — the `expectedErrors` ask was the only thing outstanding and it's in.

For the record on the undocumented-API dependency: we're comfortable taking it. The degradation path is empirically proven (your integration mock 404s every `setStatus` call, so the suite has been exercising it all along), and with `expectedErrors` in place a Slack-side withdrawal now costs nothing but the feature itself, silently.

The anchor-lingering note in DCM and the `clearTyping` contract in the platform implementation guide are both welcome as follow-ups, whenever suits.

Issue #491, opened 2026-08-21
Session-resume refusal @-mentions the bot it refuses, causing an unbounded bot-to-bot loop


theprsi:
**Version:** 1.27.0 · **Platform:** Mattermost · self-hosted, ~16 bots on one server

### Summary

The "not authorized to resume this session" refusal **@-mentions the user it is refusing**. When that user is another claude-threads bot, the refusal is itself a mention — so it wakes the bot it just refused, which replies, which triggers another refusal.

Two of our bots exchanged **1,941 messages in 37 minutes** this way. It ended only because we stopped one of the services by hand; nothing in the loop was self-limiting.

### Reproduction

1. Two bots, A and B, each on the other's `allowedUsers` (so they can tag each other).
2. Bot A has a thread session that has **timed out** (paused, still in the session store).
3. Bot B posts anything into that thread.
4. A's claude-threads sees a message in a paused thread from a user not allowed to resume it, and posts:

   ```
   ⚠️ @BotB is not authorized to resume this session
   ```

5. That message @-mentions B → passes B's `respondOnlyWhenMentioned` gate → B is invoked → B replies → back to step 4.

### Why the bot can't break out of it

Three things compound, and the third is the one I'd most like your view on:

**The model can't see what it's being asked.** `extractPrompt` strips the bot's own mention, so every one of the ~950 messages B received read as:

```
⚠️  is not authorized to resume this session
```

No subject, no addressee, nothing answerable.

**The model reasoned correctly and it didn't help.** B's replies were things like *"these are platform session-resume notices, not requests to me — nothing for me to do."* Correct every single time, and every one of them was another message that re-triggered A.

**There is no way for a session to decline to post.** B eventually emitted the literal text `(silent — per above)` — it was *trying* to stay quiet, and that attempt was published as a message. As far as I can tell there's no "produce no output" path once a session is invoked, so a bot that correctly concludes it has nothing to say still cannot stop the loop.

### Suggested fixes

1. **Don't @-mention in the refusal.** A message whose purpose is "stop talking to me" shouldn't be written in the one form guaranteed to make the recipient talk. Inline code (`` `BotB` ``) reads the same to a human and notifies nobody. This one change alone would have prevented the incident.

2. **Rate-limit the refusal** — once per (thread, user) per N minutes rather than once per message. It's the same information every time; a loop fires seconds apart and goes quiet, while a human who wanders back later still gets told why nothing happened.

[...]

PR #527, opened 2026-09-02
Slack: accept posts made through the app's own user token as the person's message

### LINKED ISSUE #526: Slack: posts made through the app's own user token never reach Claude (bot_id check)
Hi Anne — one more from the VVS fork, and this one is small: a three-line change to the Slack client's "is this a bot message" check, plus tests. PR follows today; the issue is so you have the *why* first.

**What we're building:** a live voice conversation with the agent. A person talks in the browser, a voice model (Gemini Live) acts as a front desk, and what the person asks is **posted into the task channel as that person**, using a Slack user token the person granted at sign-in. claude-threads is untouched: it just sees a message in a channel and answers as usual, and the front desk reads the reply back aloud. Code, if you're curious: https://github.com/kaza/claude-threads-upstream/tree/local/voice-desk/voice — a separate Bun service in a `voice/` directory, not something we're proposing upstream.

**What we hit:** the post never reached Claude. Slack stamps **every** API-posted message with the posting app's `bot_id` and `app_id`, even when it's posted with a person's user token. So the message arrives as

```json
{ "user": "U…person", "bot_id": "B0BUDMNPD26", "app_id": "A0BSH7L7N9E", "bot_profile": { "name": "Claude Code", … }, "text": "<@Ubot> …" }
```

and `handleSlackEvent` drops it at `if (event.user === this.botUserId || event.bot_id) return;`. Silently, which is right for that check's purpose but cost us an hour. There is no way to post a "pure" human message through the API, so any integration that speaks *for* a person — a voice desk, a phone bridge, an approval UI — is blocked by that line.

**Proposed change** (branch: https://github.com/kaza/claude-threads-upstream/tree/pr/app-user-posts, spec in `docs/app-user-posts-spec.md`):

- A message is bot-authored when its `user` is the bot itself, **or** it carries a `bot_id` and is *not* this app's with a human `user`. So: our own replies still ignored, other apps' bots still ignored, classic bot posts without a `user` still ignored. Only posts made through **this app's own user token** get through, as the person's message — same routing, same allow-list.
- The client learns its own `app_id` from the socket: `hello` carries `connection_info.app_id`, and every `events_api` envelope carries `api_app_id`. No config, no extra API call. Until it's known, the old rule applies.
- One predicate, `isBotAuthored()`, used at the three places the check lives: live events, missed-message recovery, thread history with `excludeBotMessages`.
- Why `app_id` and not `bot_id`: the `bot_id` on a user-token post is a per-authorisation bot profile, different from the bot token's own `bot_id` from `auth.test`. `app_id` is the stable identity.

**One question:** any reason you'd want this behind a config flag? With `app_id` checked, only user tokens of *this* app qualify, so we didn't see anything to opt out of — but it's your call and a one-line change either way.

kaza:
Closes #526.

Slack stamps every API-posted message with the posting app's `bot_id` and `app_id`, including one posted with a **person's user token of this app** (an integration relaying what the person said, e.g. a voice front desk). The client dropped all of them at `if (event.user === this.botUserId || event.bot_id) return;`, silently.

**Change:** a message is bot-authored when its `user` is the bot itself, **or** it carries a `bot_id` and is not this app's with a human `user`. One predicate, `isBotAuthored()`, at the three places the check lived: live events, missed-message recovery, thread history with `excludeBotMessages`.

- The client learns its own `app_id` from the socket (`hello` → `connection_info.app_id`, and every `events_api` envelope → `api_app_id`); secondary clients on a shared socket get it with each injected event. No config, no extra API call. Until it's known, the old rule applies.
- `app_id` added to `SlackEvent` / `SlackMessage` types (it was already on the wire).
- Why `app_id` and not `bot_id`: the `bot_id` on a user-token post is a per-authorisation bot profile, different from the bot token's own `bot_id` from `auth.test`.

**Still ignored, with tests:** the bot's own replies, other apps' bots, classic bot posts without a `user`, and anything with a `bot_id` before the app id is known.

Spec: `docs/app-user-posts-spec.md`. 6 new unit tests in `client.test.ts`; full suite green (3301).

### DIFF
--- a/docs/app-user-posts-spec.md
+++ b/docs/app-user-posts-spec.md
@@ -1,0 +1,74 @@
+# Posts made through the app's own user token are the person's messages
+
+Upstream discussion: anneschuth/claude-threads#TBD.
+
+## What it does
+
+A message posted with `chat.postMessage` and a **user token of this app**
+(`xoxp-…`, obtained when a person authorises the app) shows up in Slack as
+that person, and Slack delivers it with `user` = the person **and**
+`bot_id` + `app_id` of the app. Today the Slack client drops every message
+that carries a `bot_id`, so such a post never reaches Claude, silently.
+
+With this change the client treats a message as bot-authored only when
+
+- its `user` is the bot's own user (the daemon's replies), or
+- it carries a `bot_id` and is **not** from this app with a human `user`.
+
+So a post relayed through this app's own user token is handled exactly like
+the person typing it: same session routing, same allow-list, same everything.
+Other apps' bot posts, classic bot posts without a `user`, and the daemon's
+own replies stay ignored, as before.
+
+The rule is applied in the three places the filter lives: live events,
+missed-message recovery after a reconnect, and thread history with
+`excludeBotMessages`.
+
+## Why
+
+Any integration that speaks *for* a person — a voice front desk, a phone
+bridge, an approval UI — posts on the person's behalf with the app's user
+token, because that is the only way to keep attribution honest and the bot's
+consent gate (channel membership, allow-lists) intact. Slack marks every
+API-posted message with the app's `bot_id`; there is no way to post a
+"pure" human message through the API. Without this exception the daemon
+can never be driven by such an integration.
+
+Origin: the VVS live voice desk (`voice/` on the `local/voice-desk` branch
+of the kaza fork) relayed a spoken question into a task channel as the
+speaker, and the daemon ACKed the event and dropped it at the `bot_id`
+check with no log line. Seen 2026-09-02, message
+`{user: U…, bot_id: B0BUDMNPD26, app_id: A0BSH7L7N9E, bot_profile: {…}}`.
+
+## How
+
+- The client learns its own app id from the Socket Mode `hello`
+  (`connection_info.app_id`) and, belt and braces, from every `events_api`
+  envelope (`payload.api_app_id`). Secondary clients on a shared socket get
+  it handed over with each injected event. No new config, no new API call.
+- One predicate, `isBotAuthored(message)`, replaces the three inline
+  `user === botUserId || bot_id` checks.
+- `app_id` is added to the `SlackEvent` and `SlackMessage` types; it was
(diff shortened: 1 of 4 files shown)


anneschuth (approved):
Thank you — this got the full adversarial security treatment, and it holds up. **Approving.**

## Verified independently (scratch worktree on `e93d46c`)

- `tsc`/lint clean; **3302/3302 tests**.
- **Not spoofable without a real user token of this app**: the predicate requires `bot_id` + `app_id` + `team` + `user`, all stamped server-side by Slack; the identity anchors come only from server-authoritative sources (`auth.test` before the socket opens, `hello`'s `connection_info`, per-envelope `api_app_id` — Socket Mode is per-app). The secondary-injection path propagates the *parent's learned* appId, never event content.
- **Loop hazards (#491 class) all closed**: own replies dropped by the `user === botUserId` first check; a second claude-threads instance on the same app shares the bot user → dropped; same app across Slack Connect dropped by the `team` condition — and the red-check confirms that condition is genuinely load-bearing.
- **Fail-closed before identity is known**, with the test guarding the exact spoof-adjacent bug (trusting the message's own `app_id`) verified red.
- **All three filter sites** (live, recovery, `excludeBotMessages` history) use the shared predicate; no stray `bot_id` checks remain.
- **3/3 red-checks passed** — reverting the predicate fails exactly the accept-path tests while the ignore-path tests stay green.
- Mattermost parity: no analogous gap there (it only filters the bot's own id; personal-token posts already flow through), so this is a parity fix, not a new asymmetry.

## Two small non-blocking follow-ups

1. **One plain sentence in `docs/app-user-posts-spec.md` about the blast radius**: a leaked or abused user token of this app can now drive the bot *as that person*, owner-gated commands included — the same blast radius the token already had for reactions (a user token with `reactions:write` could already approve permission prompts and plans pre-PR). The spec's reasoning implies this; it deserves saying outright.
2. The PR body predates the `team` condition commit — worth a one-line refresh for the record.

Merging per the standing order.

---

Issue #383, opened 2026-05-13
hide all technical messages?


roskee:
Is it possible to not send any of the technical messages that tell me the sessions, the summary table to the mattermost channels? every time I start a new thread, claude pins this big message below it.

roskee:
I am surprised you pulled this off this fast! Thanks! I will pull the new changes and close this after testing it🙏

anneschuth:
Still going through a few tests, will release a new version soon

anneschuth:
Hi @roskee, this just landed on `main` (PR #384) and will ship in the next release.

Three ways to use it:

**Setup wizard** — `claude-threads --setup`, pick **Hidden** when asked "How verbose should the bot be in this channel?"

**CLI flag** — `claude-threads --session-header hidden --sticky-message hidden`

**YAML** (`~/.config/claude-threads/config.yaml`):
```yaml
platforms:
  - id: mattermost-main
    type: mattermost
    # ... your credentials ...
    sessionHeader: hidden    # no header table per thread
    stickyMessage: hidden    # no sticky at the bottom
```

`hidden` removes the post entirely. `minimal` keeps a one-line status bar if you want some context. `full` (default) is unchanged. Restart the bot after editing the YAML.

I'll ping back here once it's released to npm.

anneschuth:
Released in [v1.16.0](https://github.com/anneschuth/claude-threads/releases/tag/v1.16.0). Update with:

```
bun install -g claude-threads@1.16.0
```

Then either re-run `claude-threads --setup` and pick **Hidden**, or add `sessionHeader: hidden` and `stickyMessage: hidden` to your platform in `~/.config/claude-threads/config.yaml` and restart the bot.

roskee:
@anneschuth it is working as expected. Thanks 🙏

PR #529, opened 2026-09-02
Per-platform lifecycle-notice visibility (second half of #505)

kaza:
## Summary

Addresses the second half of #505 — the lifecycle status posts.

`sessionHeader` and `stickyMessage` already let an operator turn down the session table and the pinned status. Nothing touches the lifecycle notices, and in direct-channel-mode use they are most of what a quiet channel contains. This adds `lifecycle` beside them: same `full` / `minimal` / `hidden` shape, same `resolveOverheadVisibility` validation, defaulting to `full` so every existing config behaves exactly as it does today.

I've left the tool-call streaming half of #505 alone — that's a separate decision about `transformer.ts` and I didn't want to presume its shape.

## Changes

- **`lifecycle?: OverheadVisibility`** on `PlatformInstanceConfig`, resolved in `index.ts` and carried through `PlatformOverhead`. Making the field required on `PlatformOverhead` let the compiler find every construction site rather than me guessing at them.

- **`src/session/lifecycle-visibility.ts`** — one pure function, `shouldPostLifecycle(visibility, kind)`, over four named posts (`idle-warning`, `timed-out`, `paused`, `abnormal-exit`). The three-way policy is readable in one place instead of spread across four call sites in `lifecycle.ts`, and testable without a platform.

| Level | Behaviour |
|---|---|
| `full` (default) | everything, unchanged |
| `minimal` | drops only the idle warning |
| `hidden` | no status notices |

**Why `minimal` cuts specifically the idle warning:** *"Session idle - will timeout in ~N minutes without activity"* predicts something that has not happened and often never does. The timeout is `killSession(unpersist=false)` — resumable — so the next message brings the session straight back. The other three notices report a state change that already occurred, which is a different kind of message.

## One judgement call, which I'd rather you overrule than not notice

**`abnormal-exit` posts at every level, including `hidden`.**

`[Exited: <code>]` fires only for a non-zero exit code, so it is a failure report rather than overhead. Silencing it would make a session that *died* indistinguishable from one that *finished quietly* — the single case where quiet is worse than noisy, and precisely the case where someone reaching for `hidden` would least expect to lose information.

That does mean `hidden` isn't literally "post nothing", which is a defensible thing to object to. It's isolated in one branch of one pure function with a test pinning it (`never silences a failure at any level`), so it's a two-line change if you'd rather `hidden` mean absolute silence.

## Testing

- `bun run test` — 3381 pass / 0 fail; `typecheck` and `lint` clean
- New unit tests cover each level against each of the four post kinds, plus the invariant that a failure survives all three
- Running on our deployment since 2026-09-02 (at the default `full`, so as a no-op — the point being that the default path is exercised)

## Note on the other half of #505

The reporter is on Mattermost, where `sendTyping()` is implemented and gives a liveness signal independent of the stream. On Slack it was a no-op, so quieting output there would have left a channel silent while the bot worked. #525 fills that in, which makes turning things down a reasonable bargain on both platforms rather than just one. Independent of this PR, but the two are more useful together.

## Checklist

- [ ] Documentation updated (if needed) — happy to add `lifecycle` to `docs/CONFIGURATION.md` next to the other two; tell me if you want the `abnormal-exit` carve-out spelled out there too.


anneschuth (changes requested):
Thank you — the core of this is exactly right: a small, focused diff, the policy readable in one pure function, and the `abnormal-exit` carve-out is a judgement call we agree with (a dead session must never be indistinguishable from a quiet finish; keep it). We verified independently on a scratch worktree of 364c371:

**What holds up:**
- `tsc`, lint clean; unit tests **3299 pass / 0 fail**.
- Red-checks 2/2: breaking the `minimal` branch and deleting the abnormal-exit carve-out each made the corresponding tests fail — real regression tests.
- The `full` default is genuinely byte-identical to today: no post text altered anywhere, so the #491 status-post loop-protection patterns in `message-handler.ts` are unaffected. No persisted-data changes; onboarding round-trips preserve the new key; unattended routine/watch fires stay coherent (failures still surface at `hidden`).

**Two things we'd like fixed before merging:**

1. **DM auto-discovery drops the setting** (`src/index.ts` ~744). The DM-discovery `registerPlatform` closure builds `overhead: { sessionHeader: ..., stickyMessage: 'hidden' }` without `lifecycle`, and `addPlatform` takes a `Partial<PlatformOverhead>` — so the required-field compiler net you built doesn't reach this site, and DM channels silently reset to `full`. Painfully, DM channels are exactly the assistant-style use case #505 describes. `deriveDmPlatformConfig` already spreads the parent config, so it's a one-liner: pass `lifecycle: resolveOverheadVisibility(dmConfig.lifecycle, ...)` through. (The comment a few lines above warns about this exact class of bug for memory/routines/watches — it claimed another victim.)

2. **The resume notice — which #505 explicitly names — isn't gated, and `hidden` currently makes it worse.** With the pause/timeout posts suppressed, `lifecyclePostId` stays unset, so when a user messages a `hidden` thread later, `resumeSessionImpl` falls into its no-post-to-edit branch and posts a brand-new "**Session resumed** after bot restart" — misleading text (no restart happened) plus the push notification the user was trying to avoid. Net effect: at `hidden`, a pause→resume cycle posts as much as `full`, just wrongly. Our preference: add a `resumed` kind to the enum (dropped at `hidden`; keeping it at `minimal` seems right, by your own predicts-vs-reports distinction). At minimum, gate that fallback branch.

**Smaller asks, non-blocking but cheap while you're in there:**
- One wiring test: only the pure function is covered today — we removed the idle-warning gate from `cleanupIdleSessions` and the whole session suite stayed green. One test with a `hidden` overhead mock asserting no post would pin the call-site wiring, per the repo's test-the-actual-code-path rule.
- The `lifecycle` row in `docs/CONFIGURATION.md` next to `sessionHeader` — yes please, and yes to spelling out the abnormal-exit carve-out there.
- The PR body's test count says 3381; we measure 3299 on this head (base 3295 + the 4 new). Same stale-counter as #525 — worth a look at whatever produces that number.

The `minimal`-drops-only-the-idle-warning reasoning is well argued and the behavior table is exactly how this should be documented. With the two fixes above this is a clear yes, and it pairs nicely with #525 for the next release.

---

anneschuth:
Status: **two changes left**, both small.

1. `lifecycle` through the DM-discovery call site in `src/index.ts` (~744) — the one-liner off `deriveDmPlatformConfig`'s spread.
2. The resume notice: add a `resumed` kind, dropped at `hidden`, kept at `minimal`. At `hidden` today a pause→resume cycle posts *more* wrongly than it did before, which is the opposite of the feature.

The `CONFIGURATION.md` row and the wiring test are cheap while you're in there, but they're not what's holding it. The `abnormal-exit` carve-out stays as you built it — you were right to flag it and right to keep it.

kaza:
Both fixes in, plus the two smaller asks.

**1. The resume notice.** Added a `resumed` kind, dropped at `hidden`, kept at `minimal` — your read of the predicts-vs-reports distinction is right.

Only the *create* branch is gated. The other one edits the existing pause post, which adds neither a post nor a notification; gating that would leave a thread permanently showing "Session paused" for a session that resumed, which is worse than the noise. Your diagnosis of the mechanism was exact: with the pause post suppressed no `lifecyclePostId` is ever stored, so `hidden` was the *only* configuration that always took the fallback branch.

**2. DM auto-discovery.** Fixed. The comment a few lines above did warn about this exact class — and it has now claimed a victim in four PRs in this batch (#529, #534, #535, and #531, where I wrote the same bug myself by registering an event handler at startup that runtime-registered platforms never got). I have not attempted the shared-overhead-resolver refactor here since it would collide with #535 at the same call site, but the pattern is worth its own change.

**Both smaller asks, done:** the `lifecycle` row in `CONFIGURATION.md` with the abnormal-exit carve-out spelled out, and wiring tests for the call sites.

The wiring tests took two attempts, which is worth recording. The first pair passed vacuously: `resumeSession` threw on a missing `getMcpConfig` long before reaching the notice, so "no post at `hidden`" was true for entirely the wrong reason. The harness now supplies one and pins `CLAUDE_PATH` per describe — these tests reach `claude.start()`, and another test file sets `CLAUDE_PATH` to a nonexistent binary, which leaked in as ENOENT depending on file order. Both gates verified red by deleting them; your point about removing the idle-warning gate and the suite staying green now fails as it should.

The `abnormal-exit` carve-out stays as built.

3415 tests pass; typecheck, eslint and `git diff --check` clean. Reviewed by Codex and Gemini — no findings on this one; Gemini independently agreed that gating only the create branch is the right split. **CodeRabbit did not run**: its free CLI tier hit its limit partway through this batch, so this PR has two reviewers rather than three.

Also merged `main` — the branch was several commits behind by now.

kaza:
Addendum: CodeRabbit ran after all — our CLI seat was misconfigured, which is why my earlier comment said two reviewers rather than three. It came back with **0 findings across 25 files**. No changes; this one now has the full three-reviewer pass.

anneschuth (changes requested):
Thanks for the quick turnaround, and for writing up the vacuous-pass story on the wiring tests; that is exactly the kind of thing worth recording.

Both blocking asks from the first round are resolved:

- **DM auto-discovery** now passes `lifecycle` through the `addPlatform` call in `src/index.ts` (line 752 on this head). Both DM paths go through the runtime's shared `register` helper (`src/platform/dm-discovery-runtime.ts` lines 184 and 314), so live discovery and boot reconstruction carry the setting alike. Verified.
- **The resume notice** is a `resumed` kind, dropped at `hidden`, kept at `minimal`, and only the create branch of `resumeSessionImpl` is gated (`src/session/lifecycle.ts` line 1661). Gating only the create branch is the right split. The edit branch does not add a post and leaving "Session paused" standing for a resumed session would be worse.

One thing still keeps `hidden` from meaning what the docs say it means.

**1. The shutdown notice is not gated (`src/session/manager.ts` line 1748, `postShutdownMessages`).** On every graceful shutdown the bot creates "⏸️ Bot shutting down - session will resume on restart" in every active thread and stores its id as `lifecyclePostId`. Nothing in this PR touches that site. Scenario at `lifecycle: hidden`: a deploy restarts the bot, every active thread gets the shutdown post plus its push notification, and after restart the resume takes the ungated edit branch and rewrites it to "Session resumed by @user". So a restart cycle at `hidden` posts exactly as much as at `full`. For a DM assistant this is one post per DM partner per deploy, which is the noise #505 is about, and the new `CONFIGURATION.md` row promises `hidden (none)`. The fix is a few lines: the overhead map is already on the manager, so gate the `createPost` branch on `this.platformOverhead.get(session.platformId)?.lifecycle` with a `shutdown` kind (or reuse `paused`; either way `minimal` keeps it, since it reports a state change). With the create suppressed, `lifecyclePostId` stays unset and the restart resume lands in the fallback branch you just gated, so the whole cycle goes quiet. The existing edit branch should stay as is.

**2. Same shape, lower stakes: `pauseSessionsForPlatform` (`src/session/manager.ts` line 962).** "⏸️ Platform disabled - session paused" is created ungated when an operator toggles a platform off in the TUI. Rare and operator-triggered, so this is not blocking, but it is the same one-line gate.

**3. Sticky history misreads a quiet timeout (`src/operations/sticky-message/handler.ts` line 361).** `isTimedOut = isRevivable(session) && session.lifecyclePostId`. With `lifecycle: hidden` and `stickyMessage: full` a timed-out or paused session has no lifecycle post, so the history line renders it with the ✓ "completed" indicator and no resume hint, while the next message in that thread revives it. The comment right above that line describes this exact misread. The indicator should key on `isRevivable` alone; the "react 🔄 to resume" hint can stay conditional on a post existing, or say "send a message to resume" when there is none. Non-blocking, but it is the one place where this knob and #383's sticky modes interact.

Smaller notes, take or leave:

- The `abnormal-exit` gate in `handleExit` (line 2102) is unconditionally true, so the `getPlatformOverhead` lookup there can never change the outcome. Fine as a statement of intent; just noting it is not a real gate.
- No `## [Unreleased]` entry in `CHANGELOG.md`. The other feature PRs in this batch shipped with one.

What I verified on c5c3d77:

- `tsc`, eslint clean. Unit tests 3333 pass, 0 fail.
- Test-merged with current `main` (19 commits ahead of the branch): clean merge, `tsc` and the unit suite pass on the merged tree.
- The new spawn-reaching resume tests run 5 times in a row without a flake.
- Red-checks on the two wiring tests: removing the idle-warning gate and the resume gate each turns the corresponding test red.
- `full` stays byte-identical: no post text changed, the integration suites that wait on `/paused/` and `/resumed/` (`session-routines`, `session-watches`) run with the default setting and are unaffected. The Slack integration suite passes on the merged tree.
- `src/session/types.ts`'s `createResumedLifecycle` already starts a resumed session in `active`, so skipping the fallback branch at `hidden` (which was the only place `transitionTo('active')` ran for a post-less resume) has no state effect.
- Config validation goes through the existing `resolveOverheadVisibility` (throws on an invalid value at startup); the onboarding wizard preserves the key via `unmanagedFields`.

anneschuth:
Status: I merged main into this branch and pushed (`b8dae00`), because #571 had reworked the resume notice in exactly the lines this branch gates.

Worth checking my resolution, since it is the one place where the two changes met rather than sat side by side: I kept your `shouldPostLifecycle(...)` gate wrapped around **main's** new body, so the notice keeps #571's wording (the `outcome` variable and the platform-re-enabled suffix) rather than the pre-#571 text this branch was written against. `lifecycle.test.ts` and `lifecycle-visibility.test.ts` pass together, 128 cases, which is what would have caught it if I had merged the two behaviors wrongly. Say so if you would have resolved it differently.

My review round is still open on this one. Not forgotten.

anneschuth:
I pushed the three items from my second review rather than sending you another round trip (`d3554c4`), plus current main. Read it as a proposal and push back if you disagree with any of it.

**The one that mattered: `postShutdownMessages` created its post ungated** (`manager.ts:1886`). At `lifecycle: hidden` a deploy still cost a post *and* a push notification in every open thread, and because that stored a `lifecyclePostId`, the restart then took the ungated **edit** branch. A hidden thread paid full price for a restart cycle at both ends. That is the exact noise #505 exists to remove, so the doc row this PR adds was writing a cheque the code did not honor.

The create is now behind a new `shutdown` kind; the edit stays ungated deliberately, because it replaces a post the thread already has and leaving a stale "session idle" up across a restart reads worse than the edit. With the create suppressed no `lifecyclePostId` is stored, so the resume falls into the branch you had already gated. The cycle goes quiet at both ends, which is the property worth having.

`pauseSessionsForPlatform` had the same shape and got the same gate.

**The sticky bug is the one I would have missed on a re-read.** `isTimedOut` keyed on `lifecyclePostId` (`sticky-message/handler.ts:367`), so at `hidden` a paused session rendered as `✓` "completed" while the next message revives it. Your own comment two lines above says `isRevivable`, not `!cleanedAt`, for precisely this reason, and then the `&&` undid it. Now on `isRevivable` alone, and the hint names the door that actually exists: without a post there is nothing to 🔄, so it says to send a message.

**On the tests.** The `paused` and `timed-out` gates could both be deleted with the suite green, which is the failure mode you wrote up yourself earlier on this PR. Both call sites are pinned now, from both directions, matching the shape you used for the idle warning. I verified each goes red: removing the two new manager gates fails exactly the two new cases; removing the `timed-out` gate fails exactly its case.

3550 pass, lint, typecheck and knip clean.

Also: `CONFIGURATION.md` now mentions the shutdown notice, says editing is never suppressed, and warns that `hidden` removes the post 🔄 resumes from. And a `[Unreleased]` entry, which this PR was missing.

---

**One thing I am not asking you to fix here, but want on the record.** After this, #534/#535 and #547, a platform entry carries `sessionHeader`, `stickyMessage`, `lifecycle`, `toolActivity`, `toolDetails` and `turnMarker`. Nobody is going to tune six independent dials. Every participant on #505 described the same destination in words: replies only, no overhead. That is one named mode discovered three times and shipped as six orthogonal fields.

I think the right shape is a preset — something like `mode: assistant` setting the quiet combination in one line, with the individual knobs still available for people who want them. I would rather design that before the rest of the stack lands than retrofit it onto persisted user config afterwards. I will open an issue; this PR does not need to wait for it.

Approving once CI is green.

anneschuth (approved):
Green, and the three items from my second review are in (I pushed them myself as `d3554c4`; re-review welcome). 3551 pass, lint, typecheck and knip clean.

Leaving the merge itself to Anne, along with #547 and #557 — three feature PRs landing together is a release decision, not a review one. The preset question from my comment above is tracked in #590 and does not block this.

PR #530, opened 2026-09-02
cleanStale: never tombstone DCM sessions (fixes #499)

### LINKED ISSUE #499: cleanStale tombstones DCM sessions after ~1h idle — the channel goes permanently, silently deaf
## Symptom

A direct-channel-mode channel works, goes quiet for a while, the daemon restarts (deploy, reboot, crash) — and afterwards every message in that channel is silently dropped. No error, no reply, nothing in the channel. The only trace is a DEBUG line:

```
[lifecycle] No persisted session found for dcm:slac...
```

Debugged in production 2026-08-25: user messages ACKed and routed correctly, user resolved, then dropped at the persisted-session lookup.

## Root cause

Boot-time `cleanStale(sessionTimeoutMs * 2)` (manager) soft-deletes any persisted session idle longer than 2× `sessionTimeoutMinutes` — with the default 30 min, that's **one quiet hour**. The `cleanedAt` tombstone makes the session invisible to the paused-resume lookup, so the DCM channel's synthetic thread (`dcm:<platformId>`) can never resume. For thread sessions that's tolerable (a new thread just starts fresh); for DCM the *channel itself* is the conversation, so the tombstone bricks it permanently — and because the drop happens before any user-visible reply, it looks like the bot is simply ignoring people.

## Fix

Exempt DCM sessions from age-based tombstoning — their lifecycle belongs to the channel, not to an idle clock:

kaza/claude-threads@a3dbc70

(Any explicit channel teardown / archive path still ends them mechanically.)

An alternative shape, if you'd rather keep the cleanup: let the paused-resume path revive a `cleanedAt` DCM session instead of filtering it out.

kaza:
## Summary

Fixes #499. Six lines, one file.

`cleanStale` tombstones sessions that have gone quiet. A DCM session's lifecycle belongs to its channel, not to the stale sweeper, so tombstoning one makes the channel **permanently and silently deaf**: no error, no restart, no log entry a user would see. The bot simply stops answering in that channel and nothing indicates why.

## Changes

- `cleanStale` skips DCM sessions. The channel owns their lifecycle — archive/unarchive is the intended path, and the sweeper has no business ending one.

## Testing

Running on our deployment since 2026-08-25. Before the fix, a channel left alone for an hour was reliably dead on return; after it, our channels stay live across multi-hour gaps and daemon restarts. Suite green on current `main`.

## Why it matters more than its size suggests

This is the failure mode that costs the most: an agent given overnight work, and by morning the channel has been deaf for hours with no way to tell. Nothing surfaces it — you find out by discovering nothing happened.

### DIFF
--- a/src/persistence/session-store.test.ts
+++ b/src/persistence/session-store.test.ts
@@ -299,6 +299,42 @@ describe('SessionStore', () => {
       expect(history[0].threadId).toBe('old-thread');
     });
 
+    it('never tombstones a DCM session, however idle (fixes #499)', () => {
+      // A thread session and a channel session, both well past the age limit.
+      // The thread one is aged out as always; the DCM one must survive.
+      //
+      // The asymmetry is the point. A tombstoned thread session costs nothing
+      // — the next message opens a new thread and starts fresh. A DCM session
+      // IS its channel, so tombstoning it leaves nowhere else to go: every
+      // later message dies in `resumePausedSession` on "No persisted session
+      // found" (a DEBUG line, no reply), and the channel is unusable until
+      // someone edits sessions.json on the host. Idle age is the wrong owner
+      // for that lifecycle; the channel's own archive/teardown is the right
+      // one.
+      const twoHoursAgo = new Date(Date.now() - 2 * 60 * 60 * 1000).toISOString();
+      const threadSession = createTestSession({
+        threadId: 'idle-thread',
+        lastActivityAt: twoHoursAgo,
+      });
+      const channelSession = createTestSession({
+        threadId: 'dcm:slack-vvs--ch-C0BV9SBB6R2',
+        lastActivityAt: twoHoursAgo,
+      });
+
+      store.save('test-platform:idle-thread', threadSession);
+      store.save('test-platform:dcm:slack-vvs--ch-C0BV9SBB6R2', channelSession);
+
+      const staleIds = store.cleanStale(60 * 60 * 1000); // 1 hour
+
+      expect(staleIds).toEqual(['test-platform:idle-thread']);
+
+      // The channel session is still live, not merely still on disk: `load()`
+      // is what the resume path reads, and it is the lookup the bug hid it from.
+      const live = store.load();
+      expect(live.has('test-platform:dcm:slack-vvs--ch-C0BV9SBB6R2')).toBe(true);
+      expect(live.has('test-platform:idle-thread')).toBe(false);
+    });
+
     it('skips already soft-deleted sessions', () => {
       const session = createTestSession({
         threadId: 'old-thread',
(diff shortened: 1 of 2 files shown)


anneschuth (changes requested):
Thank you — the fix is correct and we independently confirmed the full deaf-channel mechanism: `findByThreadIdAnyState` deliberately includes tombstoned records (so the message routes to the resume path), while `resumePausedSession`'s `load()` skips them (so resume silently returns) — and for DCM there's no fresh-thread escape because the synthetic `dcm:<platformId>` key is the only key. That asymmetry is the deafness, your skip prevents the tombstone, and backward-compat is safe (real Mattermost post ids and Slack thread_ts can never start with `dcm:`). tsc/lint/suite all green on your head, and the diff is exactly the 6 lines described.

Two small changes before merging:

1. **Add the regression test** (repo red-green policy). We wrote it during verification to prove it's cheap — ~12 lines in the existing `cleanStale` describe block in `session-store.test.ts`: save a session with `threadId: 'dcm:test-platform'` and `lastActivityAt` 48h ago, call `cleanStale(1h)`, assert it survives and `load()` still returns it. Verified red with your fix reverted, green with it.

2. **Use the existing helper instead of the string literal**: `src/platform/utils.ts` already exports `isDcmThreadId()`, and every other consumer (8 files) imports it. Besides DRY, it's undefined-tolerant — the raw `session.threadId.startsWith('dcm:')` would throw at boot on a corrupt record missing `threadId`, which is a new failure shape on the very path that's supposed to be defensive about old data. One line: `if (isDcmThreadId(session.threadId)) continue;` (import is cycle-free).

**One thing worth a follow-up issue rather than this PR:** the boot-time sweep is only one of three tombstone entry points. `!stop` on a DCM session and a normal Claude exit both soft-delete too, and the identical lookup asymmetry then bricks the channel for up to 3 days until the permanent purge clears the tombstone. Your issue's alternative shape — reviving `cleanedAt` DCM sessions on resume — would close all entry points at once. This PR doesn't regress anything there, so we're happy to take the skip now and track the rest separately; if you want to file that issue with your production observations, even better.

(We also checked the housekeeping angle: DCM records are one-per-platform and overwritten on save, so skipping the sweep can't cause unbounded growth. No concerns.)

With the test and the helper swap, this merges immediately.

---

anneschuth (approved):
Verified the fix round on 0607f24: the new regression test is exactly what we asked for — it asserts through `load()` (the lookup the bug hid the session from), and we re-confirmed it's genuinely red with the `cleanStale` skip removed and green with it restored. tsc/lint/full suite clean, merges clean onto current main. Approving and merging.

One small thing from the review didn't make it in: the condition still reads `session.threadId.startsWith('dcm:')` rather than `isDcmThreadId(session.threadId)` (exported from `src/platform/utils.ts` — DRY, and undefined-tolerant for a corrupt persisted record). Since #538 is already reworking this file's territory, would you carry that two-line swap there rather than spinning another round here? If it doesn't fit, we'll fold it into the next release PR ourselves — either way it shouldn't hold this fix up.

Thanks for the quick turnaround — and for #538 picking up the follow-up thread this review suggested. Review of that one is in flight.

---

PR #531, opened 2026-09-02
Exit when reconnection attempts are exhausted instead of staying alive and deaf (fixes #500)

### LINKED ISSUE #500: Socket reconnect gives up permanently after max attempts — process stays 'active' but deaf
## Symptom

After a run of failed reconnects (we hit one during a Slack-side wobble: every reattempt got a disconnect request before hello, 10/10 over ~6 minutes), the client logs `Max reconnection attempts reached` — and then nothing, forever. The process keeps running, systemd reports `active`, sessions look loaded, and every message from then on vanishes. We hit this twice in one day (2026-08-24 16:54 UTC and 2026-08-25 00:52 UTC); each time a manual restart reconnected instantly.

## Root cause

`scheduleReconnect()` in `BaseClient` simply `return`s when `reconnectAttempts >= maxReconnectAttempts`. There is no terminal state: no exit, no re-arm, no external signal. For an interactively-run tool that's survivable (a human notices); for a daemon it's the worst failure mode — alive-looking and dead.

## Fix

Fail loud: when attempts are exhausted, exit the process so the supervisor (`systemd Restart=always`, launchd, Docker) revives it with fresh socket state — which demonstrably works when a stuck process's manual restart succeeds on the first try:

kaza/claude-threads@867f79c

Alternatives if exiting feels too blunt for interactive use: reset the attempt counter after a cool-down and keep trying with a capped backoff, or make the behavior configurable (`exit` | `retry-forever`).

kaza:
## Summary

Fixes #500. Five lines, one file.

When reconnection attempts are exhausted the process stayed alive but deaf — the socket was gone, nothing was listening, and systemd still reported the unit `active`. Every health check said healthy while the bot heard nothing.

## Changes

- On exhausted reconnection attempts, exit instead of continuing. `Restart=always` (or any supervisor) then revives it, which is the behaviour every deployment already expects.

## Testing

Running on our deployment since 2026-08-25. Suite green on current `main`.

## Note

This is a fail-loud change: it converts a silent, unrecoverable state into a crash a supervisor can act on. A process that cannot do its job should not report success — and "active but deaf" is the worst variety, because monitoring confirms it is fine.


anneschuth (changes requested):
Thank you — and to be clear up front: **the fail-loud direction is right and we want this fix.** "Active but deaf" genuinely is the worst failure mode for a daemon, #500 documented it twice in one day, and retry-forever alone would just hide the outage. We verified independently on a scratch worktree (tsc/lint clean, suite green, scope exactly 1 file / +5−2). The requested changes are about *where* the exit happens, not *whether*:

**1. Move the exit decision up out of `BasePlatformClient`.** An unconditional `process.exit(1)` inside the abstract client has three side effects we confirmed:
- One platform's exhaustion kills the whole daemon — healthy platforms' sessions die mid-turn with no shutdown post, bypassing the graceful path entirely. (It's also inconsistent with the repo's own startup design, where an unreachable platform is disabled and the bot continues degraded.)
- In interactive mode the reason very likely never renders: `log.error` routes through the async Ink UI, and `process.exit(1)` on the same tick means the screen just clears with no explanation. Headless/systemd gets the line; a hand-run bot gets a silent death.
- A test-run hiccup can now `process.exit(1)` the whole test runner instead of producing a diagnosable failure.

The codebase already has the template for the right shape: `!kill` exits 0 and auto-update exits 42, both *after* a graceful `prepareForRestart`-style path (persist, kill sessions cleanly, disconnect, restore terminal). Suggestion: emit a `reconnect-exhausted` event from `scheduleReconnect`, handle it in `index.ts` with that existing path, then `exit(1)`. Same supervisor semantics, one stroke fixes all three side effects, and keeps `process.exit` out of a library class.

**2. Add the regression test.** We wrote a throwaway 15-liner during verification (stub subclass, `reconnectAttempts` at max, `spyOn(process, 'exit')`): it's red on main and green on this head, so a valid red-green test provably exists and is cheap. With the event shape from ask 1 it gets even easier (assert the event fires; no exit spy needed). Repo policy is firm on this one.

**Non-blocking:**
- Your own #500 writeup offered the cool-down-reset and configurable alternatives; we don't insist on either, but if unconditional exit stays, a line in the README ("run under a supervisor; the process exits on persistent disconnection") would spare the next hand-run user a surprise.
- Worth ensuring the final error line also hits stderr directly (`console.error` before exit) so interactive users see *why* it died — with ask 1's graceful path this mostly comes for free.

For the record, we confirmed your instinct that instant exit is safe for session state: persistence is synchronous and atomic on every turn end, so `sessions.json` is current at any instant and the resume machinery handles the rest — that part needs no change.

---

anneschuth:
Correcting myself, because I've given you two different answers and you've been building against the wrong one.

On #500 I said my preferred default was never-give-up with a cool-down reset, and `reconnectPolicy: exit` as opt-in on top. My review here then asked you to move the exit out of `BasePlatformClient` without revisiting that. You've reasonably read the review as "the default is settled, fix the placement". It isn't settled, so let me settle it now.

**Ship both, with retry-forever as the default:**

```yaml
reconnectPolicy: retry   # retry (default) | exit
```

- `retry` — on exhaustion, log at error, wait the cool-down, reset the counter, keep trying. Recovers without a supervisor, which is the right behavior for the terminal-UI case.
- `exit` — what you have now. For systemd deployments like yours, where `Restart=always` is a better recovery mechanism than anything in-process.

The reason I'm not just taking your unconditional exit: this bot is run interactively often enough that a hand-run instance dying silently at 3am because the wifi dropped is a worse first impression than a noisy retry loop. And your own #500 wobble (10 failures in 6 minutes) would have self-healed under `retry` with nobody involved.

The three placement points from my review still apply to the `exit` branch — emit `reconnect-exhausted` from `scheduleReconnect`, handle it in `index.ts` through the existing graceful path, then `exit(1)`. One platform's socket must not kill sessions on healthy platforms.

That's more work than the 5 lines you filed, and I'm sorry for the detour. If you'd rather split it, a PR that only adds the `retry` cool-down (no config, no exit) would land immediately and fixes the deaf-forever bug on its own; `exit` as the follow-up.

kaza:
Reshaped to `retry` default with `exit` opt-in, and thank you for the correction — I would much rather be redirected than merge the wrong default. Your reasoning wins on its own evidence: our #500 incident was ten failures in six minutes, which `retry` would have healed with nobody involved.

```yaml
reconnectPolicy: retry   # retry (default) | exit
```

- `retry` — log at error, wait 60s, reset the counter, start a fresh round.
- `exit` — emit `reconnect-exhausted` from `scheduleReconnect`, handled in `index.ts` through the graceful path, then `exit(1)`.

All three placement points are addressed: one platform's dead socket no longer kills sessions on healthy platforms, the reason goes to stderr as well as the async Ink UI so a hand-run bot says why it died, and a test-run hiccup can no longer take out the runner. Plus the README note about running under a supervisor, and a `reconnectPolicy` row in `CONFIGURATION.md`.

What does not change is the thing #500 was actually about: returning quietly is no longer one of the options. `retry` ends the deaf state by recovering, `exit` by handing it over.

**Five defects found in review, all reproduced before fixing:**

- **Concurrent exhaustion exited mid-teardown.** `shutdown()` returned early on the second call, so its `.finally(() => exit(1))` fired immediately — before `setShuttingDown()`, persistence or session notification. Two platforms exhausting at once hit it, and so did SIGINT followed by SIGTERM, which is pre-existing but newly reachable. `shutdown()` now hands every caller the same in-flight promise.
- **A second trigger cancelled the promised cool-down.** `scheduleReconnect()` clears the pending timer at the top, so a duplicate call during the wait replaced 60s with a 1s attempt-1 backoff — measured 60,000 → 1,000ms. Slack reaches this naturally: a close before `hello` both fires `onConnectionClosed()` and rejects `connect()`, whose catch schedules again. Explicit cool-down state now short-circuits duplicates.
- **My retry test did not prove retrying.** Deleting the entire cool-down block left all five tests green — it asserted the counter reset and then cancelled the timer without ever checking a connection was attempted. Rewritten to advance time and assert `connect()` runs, plus cases for a duplicate trigger and for disconnect. All three fixes are mutation-checked.
- **A pending cool-down survived `disconnect()`**, holding the event loop open through shutdown. `disconnect()` now cancels it. Asserted on the timer itself rather than on connection attempts — the reconnect callback already bails on an intentional disconnect, so that assertion could not tell a cancelled timer from a live one.
- **The startup window swallowed exhaustion.** Before `main()` wires the shutdown hook there is nothing to shut down gracefully, and my first version logged and returned — leaving precisely the alive-and-deaf process the policy exists to prevent. It exits.

And one I wrote and caught myself: the handler was registered by looping over `platforms.values()` once at startup, but DM auto-discovery registers platforms at runtime through `wirePlatformEvents`. A derived DM platform with `reconnectPolicy: exit` would have emitted to nobody. That is the fourth time in this batch that a per-platform setting missed the DM-discovery path.

Scope is no longer 1 file / +5−2 — it is closer to 11 files. If you would still rather split it (cool-down only first, `exit` as a follow-up), say so and I will cut it; the cool-down half stands alone cleanly.

3414 tests pass; typecheck, eslint and `git diff --check` clean. Reviewed by Codex and Gemini. **CodeRabbit did not run** — free CLI tier exhausted partway through this batch.

kaza:
Addendum — CodeRabbit ran after all (our CLI seat was misconfigured earlier, which is why my last comment said two reviewers). Two majors, both real:

**1. `reconnect-exhausted` fired once per trigger, not once per round.** With `exit` the attempt counter is deliberately not reset, so every later `scheduleReconnect()` re-entered the exhausted branch and asked for shutdown again — and Slack produces two triggers per round by itself, the same close-before-`hello` path that caused the cool-down bug. Latched now, and the latch clears on a successful connection or an explicit reconnect prep, so a socket that recovers and later dies again is still reported.

**2. A secondary Slack client had a `reconnectPolicy` that could never fire.** On a shared event source a secondary never opens a socket (`connect()` registers with the parent instead), so it cannot exhaust reconnection — but it was still calling `setReconnectPolicy` from its own config, which reads as configured while doing nothing. The value is still validated for every instance so a typo stays a startup error; it is only *applied* when the client owns a socket, and a secondary whose explicit policy differs from its parent's now says so rather than being silently dropped.

That second one is the same shape as the DM-discovery findings across this batch: a per-platform setting that looks wired and isn't. Worth noting the shared-socket case is invisible to the compiler in a way the DM one wasn't.

Two regression tests, verified red. 3416 tests pass; typecheck, eslint and `git diff --check` clean. Three reviewers on this one now.

kaza:
One more round — and this one was a real bug I had introduced with the cool-down itself.

**A pending cool-down outlived a reconnection that succeeded.** If the socket came back by another route while the 60s cool-down was still waiting (a heartbeat-driven attempt landing first), the timer fired on a **healthy** connection — and `scheduleReconnect()` force-closes before reconnecting. So recovering from an outage killed the connection that had recovered it. `onConnectionEstablished()` now clears the pending reconnect before anything else.

Worth noting how that one hid: the cool-down only exists on the `retry` path, and every earlier test drove exhaustion straight through to the next attempt. Nothing exercised "exhausted, then rescued from outside" until a reviewer asked what cancels the timer.

Also folded in: two tests started a heartbeat via the connection-established path and never stopped it, so they leaked an interval.

**Not taken** — replacing the wall-clock waits in those tests with injected timers. The behaviour under test *is* timer sequencing, the waits are 20–60ms against a 1ms backoff, and an injected scheduler would largely test the injection. Happy to change it if you'd rather; noting it as a deliberate skip rather than an oversight.

3417 tests pass on the branch, 3648 on our integration branch; typecheck and eslint clean.

anneschuth (approved):
Thanks for taking the detour with the default and for reproducing each of the follow-up bugs before fixing them. Every ask from the first round is addressed, and I traced the reconnect state machine for the races the rework could have opened. Nothing blocking. One merge chore and a few non-blocking notes below.

**Merge chore: the branch conflicts with `main`.** The merge commit on this branch brought in a stale copy of `main` (the merge base is still 1.33.1), so `src/index.ts` now conflicts at the `const shutdown` line. It is a trivial one: main added `void session.noticeClaudeAiConnectors();` right before `const shutdown`, your side rewrote `shutdown` into the `shutdownInFlight` wrapper. Keep both. Merging current `main` also picks up #556 (the shutdown flag set before the first await), which your `runShutdown` should preserve as-is.

**Non-blocking**

1. `src/index.ts:1051-1058`, the `exit` path has no deadline. `runShutdown` awaits `updateAllStickyMessages()`, `postShutdownMessages()` and `killAllSessions()`, all of which make HTTP calls with retry-with-backoff to the platform that just proved unreachable for ten attempts. When the whole network is down (not a Socket Mode wobble) each call burns through its retries before failing, so the process sits in "shutting down" for minutes before `process.exit(1)` fires. That is a bounded delay, not a hang, but it is the alive-and-deaf state with a banner on it. A `Promise.race` against a 30s timer around the `shutdown()` call in `onReconnectExhausted` would keep the exit honest. Sessions are persisted on every turn end, so a forced exit loses nothing.

2. `src/index.ts:885-887`, `prepareForRestart` sets `isShuttingDown` but does not set `shutdownInFlight`. If a platform exhausts while an update restart is mid-teardown, `shutdown()` calls `runShutdown()`, which returns at the `isShuttingDown` guard, and `.finally(() => process.exit(1))` fires while `killAllSessions()` and the disconnects are still in flight. Under the bundled daemon or `Restart=always` the process comes back either way, so the damage is a lost update restart and an unclean session kill. Same shape as the SIGINT-during-update case you noted as pre-existing; the fix is to route `prepareForRestart` through the same in-flight promise.

3. `src/platform/base-client.ts:446`, `scheduleReconnect` still has no `isIntentionalDisconnect` guard at the top. Attempt 10's `connect()` in flight, then `disconnect()` (platform toggle or shutdown), then the rejection lands: `clearReconnectTimer()` already ran, `reconnectAttempts` is still at max, so the `retry` branch logs "exhausted, retrying in 60s" and arms a cool-down on a client that was deliberately closed. The eventual attempt bails on the existing intentional-disconnect check inside the timer, so this is a spurious error line plus a 60s timer, visible only on the disable toggle since the shutdown path exits explicitly. An early `if (this.isIntentionalDisconnect) return;` closes it.

4. `docs/CONFIGURATION.md`, Slack table. The row does not say that a channel sharing another entry's Socket Mode connection ignores its own `reconnectPolicy` (the constructor warns at startup, `src/platform/slack/client.ts:123-131`). One clause in the row would save someone setting it on the wrong entry.

5. During the 60s cool-down the UI keeps showing "reconnecting (attempt 10/10)" because nothing is emitted between the last `reconnecting` event and attempt 1 of the next round. A `reconnecting` emit with the cool-down, or a log line the headless provider shows, would make the wait legible.

**Verified**

- First-round asks: the exit decision moved out of the client (`emit('reconnect-exhausted')` at `base-client.ts:471`, handled in `index.ts:128-141` and `index.ts:1051-1058` through `shutdown()` then `exit(1)`); the reason reaches stderr via `console.error` as well as the Ink log; no `process.exit` left in library code; regression tests in `src/platform/reconnect-exhausted.test.ts` (the exit-event test is red on `main`, which never emits) and `src/config/reconnect-policy.test.ts`; README "Staying connected" section and the `reconnectPolicy` rows in both config tables.
- The settled default: `resolveReconnectPolicy` returns `retry` for undefined and null, validates `retry` and `exit` with the field path in the error, and both platform constructors call it, so a typo is a startup error. `deriveDmPlatformConfig` spreads the parent config, so DM-derived instances inherit the policy.
- Races, by reading `scheduleReconnect`, `onConnectionEstablished`, `onConnectionClosed`, `disconnect` and both platforms' `connect()`: the two triggers per failed attempt (Slack close-before-hello and Mattermost error-then-close both do this) are absorbed by `cooldownActive` on the retry path and `exhaustedEmitted` on the exit path; the heartbeat only runs after `onConnectionEstablished`, which now clears the cool-down first, so a heartbeat can never fire into a pending cool-down; a connection that succeeds and closes immediately resets both the counter and the latch and starts a fresh round, so a later exhaustion is reported again; `disconnect()` cancels the cool-down; two platforms exhausting at once share one `runShutdown` through `shutdownInFlight`, and whichever caller registered first decides the exit code, so SIGINT before exhaustion still exits 0 and the daemon does not restart.
- Daemon wrapper: `--auto-restart` passes `--restart-on-error` to `bin/claude-threads-daemon`, so exit 1 restarts there as it does under `Restart=always`; a bare `claude-threads` run with `exit` configured stays down, which is what the README now says.
- On the branch head: `tsc`, `eslint` and `knip` clean; the 107 tests across the two new files, `shared-event-source.test.ts`, `slack/client.test.ts` and `mattermost/client.test.ts` pass. Timer sequencing in the new tests is safe under load since every assertion runs on a timer due before the one it checks against.

PR #532, opened 2026-09-02
Count Slack protocol pings as socket activity (fixes #498)

### LINKED ISSUE #498: Slack Socket Mode: 60s heartbeat kills healthy idle connections (protocol pings are invisible to onmessage)
## Symptom

On an idle workspace, the Slack client reconnect-loops forever: every ~60s the heartbeat logs `Connection dead (no activity for 60s), reconnecting...`, the reconnect succeeds, hello arrives, then 60 idle seconds later it's executed again. Messages sent during the ~2s reconnect gaps are lost (Socket Mode has no replay). Observed continuously on our production deployment (Bun 1.4.0, AL2023 arm64) from 2026-08-25 02:28 UTC until patched.

## Root cause

`BaseClient`'s heartbeat counts activity only via `updateLastMessageTime()`, which is fed exclusively from `ws.onmessage` — i.e. text frames. An idle Slack Socket Mode connection sends **no text frames at all** after `hello`; it keeps the connection alive with **protocol-level ping frames every ~10s**, which never reach `onmessage`.

Measured with a raw `WebSocket` against `apps.connections.open` (150s idle window):

```
0.8s  OPEN
0.8s  FRAME {"type":"hello",...}
9.8s  PING-EVENT
20.0s PING-EVENT
30.3s PING-EVENT
... every ~10.2s, no text frames ...
```

So on a quiet workspace the 60s timeout (`HEARTBEAT_TIMEOUT_MS`) always fires against a perfectly healthy socket.

## Fix

Bun's WebSocket surfaces those ping frames as nonstandard `'ping'` events. Subscribing to them and feeding the activity clock fixes it — one listener next to the `onmessage` wiring:

kaza/claude-threads@144e2d0

Verified: previously 1 kill/minute; after the patch, hours idle with zero false deaths.

Note: browser/Node `ws` environments may need a different source of ping visibility (`ws` package exposes a `'ping'` event too); an alternative is client-initiated pings from the heartbeat.

kaza:
## Summary

Fixes #498. Eight lines, one file.

The 60-second heartbeat only counted text frames as activity. An idle Slack Socket Mode connection sends only protocol pings (~10s cadence), which the handler never saw — so the heartbeat concluded every healthy idle connection was dead and killed it, once a minute, forever.

## Changes

- Count Slack protocol pings as socket activity.

## Testing

Verified against live Slack: 5+ idle minutes with zero false deaths, where before the connection was torn down every 60 seconds. Running on our deployment since 2026-08-25. Suite green on current `main`.

## Note

The reconnects were mostly invisible because they succeeded — the cost was a permanent churn of teardowns and reconnects on every idle channel, and each one is a window in which a message can be missed.

### DIFF
--- a/src/platform/slack/client.ts
+++ b/src/platform/slack/client.ts
@@ -1,4 +1,4 @@
-import { WebSocket } from '../../utils/websocket.js';
+import { WebSocket, countPingsAsActivity } from '../../utils/websocket.js';
 import type { SlackPlatformConfig } from '../../config/index.js';
 import { wsLogger, createLogger } from '../../utils/logger.js';
 import { truncateMessageSafely, escapeRegExp, getEmojiName, formatWebSocketError, resolvePostThreadId, isDcmThreadId, normalizeAckReaction, resolveDirectChannelMode, type ResolvedDirectChannelMode, type ApprovalsMode } from '../utils.js';
@@ -460,12 +460,11 @@ export class SlackClient extends BasePlatformClient {
       };
 
       // Slack keeps idle Socket Mode connections alive with protocol-level ping
-      // frames (~10s cadence) — they never reach onmessage. Bun's WebSocket
-      // surfaces them as 'ping' events; count them as activity, or the 60s
-      // heartbeat executes every healthy idle connection (observed 2026-08-25).
-      (this.ws as unknown as EventTarget).addEventListener?.('ping', () => {
-        this.updateLastMessageTime();
-      });
+      // frames (~10s cadence) — they never reach onmessage. Without counting
+      // them the 60s heartbeat executes every healthy idle connection
+      // (observed 2026-08-25). Covers Bun and Node 20; see the helper for why
+      // Node 22+/undici cannot be covered here (#498).
+      countPingsAsActivity(this.ws, () => this.updateLastMessageTime());
 
       this.ws.onmessage = (event) => {
         this.updateLastMessageTime();
--- a/src/platform/slack/client.ts
+++ b/src/platform/slack/client.ts
@@ -459,6 +459,14 @@ export class SlackClient extends BasePlatformClient {
         wsLogger.info('Socket Mode: WebSocket connected, waiting for hello...');
       };
 
+      // Slack keeps idle Socket Mode connections alive with protocol-level ping
+      // frames (~10s cadence) — they never reach onmessage. Bun's WebSocket
+      // surfaces them as 'ping' events; count them as activity, or the 60s
+      // heartbeat executes every healthy idle connection (observed 2026-08-25).
+      (this.ws as unknown as EventTarget).addEventListener?.('ping', () => {
+        this.updateLastMessageTime();
+      });
+
       this.ws.onmessage = (event) => {
         this.updateLastMessageTime();
 
--- a/src/utils/websocket.test.ts
+++ b/src/utils/websocket.test.ts
@@ -1,0 +1,97 @@
+/**
+ * Regression tests for the ping-as-activity attachment (#498).
+ *
+ * An idle Socket Mode connection carries only protocol-level ping frames, so
+ * a heartbeat fed from `onmessage` alone concludes every healthy idle
+ * connection is dead and kills it once a minute, forever. These pin both the
+ * fake-object contract and the real behavior against a live `ws` server.
+ */
+
+import { describe, it, expect } from 'bun:test';
+import { WebSocketServer } from 'ws';
+import type { AddressInfo } from 'node:net';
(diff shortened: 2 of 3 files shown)


anneschuth (changes requested):
Thank you — the root cause is exactly as your #498 trace showed, and we went further and verified the fix **empirically across runtimes**: a local `ws` server sending protocol pings every 150ms against a client using your exact attachment shape, under Bun's native WebSocket, Node 22's undici global, and Node 20's `ws` package (the repo's compat layer in `src/utils/websocket.ts` uses all three; CI's node-smoke matrix runs the built binary under Node 20/22/24). Results:

| Runtime | `addEventListener('ping')` | Throws? |
|---|---|---|
| Bun (native) | **fires, 7/7 pings** — bug fixed | no |
| Node 22+ (undici) | silent no-op — #498 remains | no |
| Node 20 (`ws` pkg) | silent no-op (`ws`'s EventTarget shim only wires open/message/close/error) | no |

So the fix is real and completely safe (the optional-chained cast can't throw anywhere), but it's **Bun-only** — under both Node paths the heartbeat still kills healthy idle connections. Two asks before merging:

1. **Cover Node 20 with one more line** — `ws`'s EventEmitter interface does emit `'ping'`, and we verified this works against `ws@8.21.3`:
   ```ts
   (this.ws as unknown as { on?: (ev: string, fn: () => void) => void }).on?.('ping', () => this.updateLastMessageTime());
   ```
   Safe no-op on Bun/undici (neither has `.on`). And please note in the comment (or PR body) that Node 22+/undici remains uncovered — undici auto-pongs internally and exposes no frame-level ping visibility at all, so that gap is a dead end at this layer, and #498 shouldn't be closed as fully fixed for that runtime. (Client-initiated pings from the heartbeat, your own suggested alternative, would be the eventual undici answer — follow-up material, not this PR.)

2. **A regression test** (repo red-green policy). It's feasible in-repo: `ws` is already a dependency, so a test can spin up a local `WebSocketServer` that calls `ws.ping()` and assert the activity clock advances — or extract the listener attachment into a tiny helper and test it against a fake emitter. Bun-only coverage is fine since the suite runs under Bun.

Everything else checked out clean: exactly 1 file / +8 lines, correctly Slack-scoped (Mattermost feeds its clock from `onmessage`, untouched), no interaction with #502's shared event source (secondaries never own a socket, never start a heartbeat), no listener leak across reconnects, and a verified conflict-free merge onto current main despite being one commit behind. The comment's claims (~10s cadence, pings invisible to `onmessage`, Bun surfacing) all verified accurate.

With the Node 20 line and a test, this merges straight away.

---

anneschuth:
Status: **one line plus a test.**

```ts
(this.ws as unknown as { on?: (ev: string, fn: () => void) => void }).on?.('ping', () => this.updateLastMessageTime());
```

That covers Node 20's `ws` path; safe no-op on Bun and undici. Plus the regression test (local `WebSocketServer` calling `ws.ping()`, assert the activity clock advances).

One thing to carry into the PR body: **Node 22+/undici stays broken** and can't be fixed at this layer — it auto-pongs internally and exposes no frame-level ping visibility. So #498 shouldn't close as fully fixed. Client-initiated pings from the heartbeat are the eventual answer there, and that's a separate PR I'd take.

anneschuth:
Pushed the Node 20 line and the regression test, same reasoning as #525 — you have a lot open and this was small.

It grew slightly past the one-liner I suggested. Since the two runtimes need different mechanisms, I pulled it into `countPingsAsActivity` in `src/utils/websocket.ts` — that module already exists to hold the runtime split, and it makes the behavior testable without a socket. The Slack client is now one call.

Both halves are independently red-checked: removing `addEventListener` fails the Bun test, removing `.on` fails the Node 20 test. There's also a live `ws`-server test, but it stays green either way (Bun's `ws` client satisfies both styles), so it's end-to-end coverage rather than a discriminator and the comment says so.

The undici gap is documented in the helper rather than left to be rediscovered, and I've noted on #498 that it shouldn't close as fully fixed for that runtime. Client-initiated pings from the heartbeat — your own suggestion — remain the answer there, and I'd take that PR.

3300 tests pass, tsc and eslint clean.

anneschuth (approved):
Approving — Node 20 is covered and the regression test is in.

Worth restating for whoever reads this later: **#498 should not close as fully fixed when this merges.** Node 22+/undici still kills healthy idle connections, and cannot be fixed at this layer — it auto-pongs internally and exposes no frame-level ping visibility at all. That's documented in the helper and noted on the issue. @thejdubb02's report is on Node 22, so their case specifically remains open.

Client-initiated pings from the heartbeat — your own suggested alternative — are the answer there, and that PR would be welcome.

PR #534, opened 2026-09-02
Tool activity: summary line or hidden, details in a thread (#505, tool half)

kaza:
The tool-call half of #505, in the shape proposed there. Refs #505 (the lifecycle half is #529).

Two per-platform fields beside `sessionHeader` / `stickyMessage`:

```yaml
toolActivity: summary   # full (default) | summary | hidden
toolDetails: thread     # thread (default for summary) | none
```

| `toolActivity` | The reply |
|---|---|
| `full` | unchanged: every tool inline with its `↳` indicator |
| `summary` | one live line at the top of the turn's post, `🔧 12 tools · 40 s`, updated as tools run; `· 1 ❌` when a tool failed; Claude's text follows |
| `hidden` | nothing about tools at all (what was agreed in #505) |

`toolDetails: thread` posts the full stream as replies in a thread under the turn's post (in a thread-mode session, after the reply in the same thread, since Slack has no nested threads). Permission prompts, plan approvals, questions, task lists and errors post as before in every mode. Defaults keep today's behaviour exactly.

**How**

- The transformer emits `tool_activity` ops (`start` / `end` / `turn_end`) instead of inline tool content in the quiet modes; `full` is untouched. Legacy top-level `tool_use` / `tool_result` events follow the mode too.
- A `ToolActivityExecutor` keeps the per-turn counter and feeds a details sink (`thread` or `none`).
- The content executor renders the summary as a header on the turn's post through its one write path, so a split can't freeze a stale count into one post and a fresh one into the next. Length checks and truncation account for the header.
- The thread sink is a second `ContentExecutor` with no task-list bump callbacks, posting as `tool_details`, never touching the latest-message pointer, and only once the reply post exists.
- Config errors at startup with the field path: `toolDetails` with `full`, `thread` with `hidden`, unknown values. Derived DM platforms inherit the parent's settings.

Spec with the plan- and code-review trail: `docs/quiet-tools-spec.md`. `docs/CONFIGURATION.md` documents the fields. 37 new tests; full suite green (3,336).

A second, stacked PR adds `toolDetails: file` (one HTML page per turn in a directory, optional URL for the summary line), which is what we run; kept apart so this one stays small.


anneschuth (changes requested):
Thank you — this is a substantial, carefully engineered feature, and it earned the two full independent verification passes we gave it (this repo's policy for large PRs). The core held up impressively: 3330 unit tests green, the full Slack integration suite green (136/136), six red-checks across both passes all genuinely red, `full` mode verified byte-identical branch by branch, the summary header's split arithmetic sound, all special ops (task lists, questions, plan approvals, subagents) flowing untouched in every mode — and the DM-discovery inheritance correctly wired this time. Subagent sidechains neither double-count nor go missing; no new watch-trigger or loop-protection exposure; memory is pruned like `full` mode.

The second pass found two real bugs, both small, both confined to files this PR touches:

**1. The reset you wrote is unreachable in production.** `ToolActivityExecutor.reset()` (and via it the sink reset) is only called from `MessageManager.reset()`, which has no non-test caller. The respawn path (`restartClaudeSession` for `!cd` / `!permissions interactive`) calls `clearClaudeSessionState()`, which clears `toolStartTimes` and the task tracker but not the tool-activity executor or thread sink. Issue a `!cd` mid-turn in summary mode and the dead turn's stats bleed into the next turn's header (inflated count, elapsed measured from the old turn's start, possibly a stuck "…"), and the sink's captured context threads the next turn's details under the *previous* turn's reply. You clearly built `reset()` for exactly this — please wire it into `clearClaudeSessionState()` and the resume-respawn path.

**2. A stale header survives a post split.** `flush()` handles `headerDirty` only in the empty-pending branch, and `flushPending` clears the header only when writing to the header post itself. Sequence: reply splits into a continuation post → last tool ends (header goes final, dirty) → Claude's closing text arrives → the result flush writes to the *continuation* post → the header post keeps the stale running header forever (we reproduced it: post 1 permanently shows `🔧 1 tool · 1 s…` instead of the final count). Your split test only covers the empty-pending case. Fix sketch: after `flushPending`, if `headerDirty` and the current post isn't the header post, re-render the header post — plus a red-green test for the split-then-closing-text sequence.

**Doc asks (cheap, please bundle):**
- `docs/quiet-tools-spec.md` opens with a YAML example using `toolDetails: file` + `toolDetailsDir`/`toolDetailsUrl` — which **this** PR rejects at startup (they're #535's). A copy-paste of your own spec's example errors out; please trim the spec to shipped scope or mark the file-sink parts as PR-2. The spec also carries your org's internal context (domain, deployment anecdotes) and references commits that don't exist in this repo — worth a pass. `CONFIGURATION.md` itself is accurate.
- A `## [Unreleased]` CHANGELOG entry, per repo convention.

**Notes, no action needed:** your branch is one commit behind main; the merge is verified conflict-free (`git merge-tree` clean, zero file overlap with #527 — and none with your #525 either), though a rebase before merge is cheap insurance since CI ran without it. The `· details` link part of `renderToolSummary` is currently unreachable (`link()` hard-codes null) — fine as the #535 hook, just noting it. And the test count in the body (37/3336) measures as +28/3330 — the same stale counter as your other PRs; whatever generates those numbers deserves a look.

With the two fixes and the doc pass, this merges — and together with #529's fixes it completes #505, which will be a genuinely good release.

---

anneschuth:
Status: **two bugs plus a doc pass**, unchanged from my review — the unreachable `reset()` and the stale header after a split.

One addition worth folding in while you're here: @thejdubb02 asked in #505 for the summary line to carry the **last tool name** (`🔧 12 tools · 40 s · Bash`), and argued it well — once the stream is hidden, that line is the only liveness signal, so it should show what the bot is doing, not just how much it has done. Agreed. Cheap to add now, and it makes the mode meaningfully better.

Also: your `reset()` finding here has a twin in #547, which adds a turn counter to the same unreachable method. Whatever you do about `clearClaudeSessionState()` should probably fix both at once.

This plus #535 plus #529 closes #505, which is the release I want to cut.

kaza:
Both fixes in, plus the doc pass and the last-tool-name addition. Three commits: the merge of `main`, the fixes, and a round-2 commit from our own review.

**1. The unreachable reset.** You were right about the cause and, it turned out, righter than my first fix. New `MessageManager.clearTurnState()`, called unconditionally by `restartClaudeSession` — resume or not, since the turn belongs to the process being killed either way — and by `clearClaudeSessionState()`, which the two hand-rolled worktree respawns use instead of the helper.

That fixed the counter. It did **not** fix the second half you described — the sink threading the next turn's details under the previous turn's reply. `turnOpen` / `headerPostId` / `headerBody` live on the content executor and clear only on a `result` flush, which a respawn never produces, so the next turn's summary still edited the abandoned post and the fresh sink took that post as its root. Same bug, one layer down. `ContentExecutor.abandonHeaderTurn()` releases it; the dead post keeps whatever header it last rendered, on the same principle as the file sink's abandoned page.

**2. The stale header after a split.** Fixed as sketched, but by the flag rather than by comparing post ids: every path that writes the header post clears `headerDirty`, so a flag still set after `flushPending` means the header post was untouched — which covers the continuation case, all three `handleSplit` early returns and the multi-chunk loop, without a check at each site. Red test for your exact sequence (split → last tool ends → closing text → result flush lands on the continuation).

That change turned out to have teeth of its own. `headerBody` advanced only on a *successful* update, so a lost response — platform accepted, client threw — left it stale, the split's continuation then cleared the pending content, and the new re-render overwrote the post with the older body. Reproduced: the post rolled back to `🔧 1 tool · 1 s\n\npara one`, losing the entire first part. `tryUpdatePost` now records the *attempted* body on failure. An update replaces the whole post, so restoring the attempt is correct whether or not it arrived; restoring the last confirmed body is not.

**3. The last tool name.** Agreed, and thank you for relaying the argument — @thejdubb02 is right that it's the only liveness signal once the stream is hidden. `🔧 12 tools · 40 s · Bash`, the most recently started tool, after the seconds. MCP names shorten to the tool part (`mcp__playwright__browser_navigate` → `browser_navigate`). Two existing tests asserted the summary line contains no tool name; both now assert the absence of the tool *line* instead, which was their actual intent.

**Docs.** The spec's opening example no longer uses the PR-2 fields this PR rejects, the file-sink rows are marked as such, and the org-internal paths, domain, deployment anecdote and the reference to a commit outside this repo are gone. Q-001 is recorded as decided rather than open. `CONFIGURATION.md`'s summary-line format now matches. CHANGELOG entry under `## [Unreleased]`.

**Also.** Merged `main` (the branch was several commits behind by now, not one; clean merge). And a defect of my own making: `ToolActivityExecutor.reset()` is reached twice on a fresh-session respawn, and #535's file sink bumps its turn number on reset — so the second call would have skipped a page the moment the stack landed. `reset()` is now a no-op when the turn has no tools to abandon.

**On the reset twin in #547** — agreed, and it should be simpler there now: `clearTurnState()` is the seam, so the turn counter belongs next to these two rather than in `MessageManager.reset()`. I'll wire it that way when I touch #547 unless you'd rather it moved here.

Six regression tests across the two commits, each verified red before its fix, and both P2 fixes re-verified by reverting them in place afterwards. 3450 tests pass; typecheck, eslint and `git diff --check` clean.

Reviewed by Gemini (no findings), CodeRabbit (no findings) and Codex, which found the two P2 defects above. One CodeRabbit finding — that `sink.append` runs before the parent post exists — was rejected: `thread.ts` queues lines for exactly that case and drains once a root appears, and the existing `summary + thread` test already asserts the reply post is created first. It did not reproduce on a second pass.

Ready for another look. #535 next, unless you'd rather I hold until this one settles.

anneschuth (changes requested):
Thanks for the thorough second round; both fixes from my review are in, the `abandonHeaderTurn()` follow-through was the right call, and the last-tool name reads exactly as @thejdubb02 argued for.

Two things still block the merge: one new defect that the round-2 "attempted body" change introduces, and the branch no longer merges.

**1. A genuinely failed update to the header post now duplicates the paragraph (round 2 regression).** `src/operations/executors/content.ts`, the `catch` block of `tryUpdatePost` (the `headerBody = content` on failure). The Codex fix assumes the failed write landed, which is right for the split path, where the continuation post clears the pending content and the header re-render is the only chance to restore the first part. It is wrong for the plain update path, where `onFailure` nulls `currentPostId` but leaves the pending content in place, so the next flush re-posts it. Sequence, summary mode: post 1 holds `para one`, a tool ends (header final, dirty), `para two` arrives, the flush's `updatePost(post_1, "para one\n\npara two")` fails with a 5xx that really was a rejection. `headerBody` now holds the attempted body, the post-flush `renderHeaderOnly` writes `🔧 1 tool · 1 s\n\npara one\n\npara two` to post 1, and the result flush creates post 2 with `para two` from the still-pending content. I reproduced it against the real executor: `para two` appears twice, in post 1 and post 2. With the catch-block line removed it appears once, and your rollback test fails, so the two scenarios genuinely conflict inside the helper. The helper's own doc comment warns against baking one state policy into it, and that is the fix: record the attempted body only at the call site whose failure drops the pending content (the split's first-part write), and keep the confirmed body where the caller will re-post the pending text (the plain update and the three `handleSplit` early returns, which also leave pending intact). A red test for the sequence above, please.

**2. The branch conflicts with main.** The merge commit brought in the 1.33.1 tip; main is at 1.35.0 plus one fix, nineteen commits past your merge base, and GitHub reports the PR as conflicting. The conflicts are in `CHANGELOG.md` (your entry belongs under `## [Unreleased]`, above the new 1.35.0 heading) and the import list in `src/index.ts` (keep both `resolvePlatformMcpPosture` and `resolveToolActivity`). I merged main locally with those two resolutions and the result passes typecheck, lint and 3490 unit tests, so it is mechanical, but the PR's CI has never run against the current main, and #535 is stacked on this branch, so it inherits the same base. Please rebase or merge the actual tip.

Two smaller ones, no blocker:

- `src/operations/tool-details/thread.ts:38`: the details executor uses its own 500 ms timer, and nothing cancels it. `!stop` flushes the main `MessageManager` (which cancels its own timer) and kills the process, so a details post can still be created or edited up to half a second after the session ended. Cancelling it from `reset()`, or flushing the sink in the kill path, would close that.
- After a `!cd` mid-turn the abandoned post keeps the header the pre-respawn flush rendered, including the running `…`, forever. You chose that deliberately and I can live with it, but a final render before `abandonHeaderTurn()` is one platform call.

What I verified:

- Ask 1 (unreachable reset): `restartClaudeSession` calls `clearTurnState()` unconditionally (`src/operations/commands/handler.ts:139`), `clearClaudeSessionState()` calls it too (`src/operations/message-manager.ts:1425`), so both hand-rolled worktree respawns (`src/operations/worktree/handler.ts:585`, `:766`) reach it, and both plugin respawns go through `restartClaudeSession`. `abandonHeaderTurn()` (`content.ts:726`) releases the header post, so the next turn's summary and details thread no longer land on the dead reply. The `handler.test.ts` and `message-manager.test.ts` tests drive the real functions.
- Ask 2 (stale header after a split): the post-`flushPending` re-render in `flush()` covers the continuation post and all three `handleSplit` early returns. I removed that render and re-ran your split-then-closing-text test: it fails, so it guards the fix.
- Last tool name: `renderToolSummary` appends the most recently started tool, MCP names shortened to the tool part, with tests.
- Docs: the spec's opening example no longer uses the PR-2 fields, the file-sink rows are marked, the internal context is gone; `CONFIGURATION.md` matches the line format; the CHANGELOG entry is under Unreleased.
- Default path: `full` / `none` keeps `toolsInline` true, no `ToolActivityExecutor`, header null, so `renderFor` returns the body unchanged and `headerReserve` is 0. Byte-identical.
- Interrupt: the real CLI emits a `result` (`error_during_execution`) on `!escape`, so the turn closes and the counter resets; no respawn needed.
- Resume after a bot restart builds a fresh `MessageManager` with the platform's tool settings; nothing new is persisted. DM platforms spread the parent config, so they inherit the mode.
- On the branch as pushed: typecheck, lint, knip and 3368 unit tests pass (the description says 3450; same stale counter as before).

anneschuth:
Status: merged main into this branch (`3e0fdea`) so it is no longer stale.

The substantive news is on #535 rather than here. The header-duplication defect from my last review is still present in both branches, and since #535 carries all of this branch's code, I pushed the fix there (`7a3365b`) with a red-green test rather than sending you another round trip. Details in that thread, including a correction to something I got wrong in my review here: moving the assignment to the split call site satisfies both your rollback test and the new repro, so the two scenarios never actually conflicted.

Which raises the question of whether this PR should stay open at all, given #535 contains it entirely. My inclination is to close this one in favor of #535 once that lands, but it is your stack, so tell me if you would rather keep them separate.

PR #535, opened 2026-09-02
Tool details: one HTML page per turn, optional URL on the summary line (stacked on #534)

kaza:
Stacked on #534 (the first four commits are that PR; review the last four). Refs #505.

Adds `toolDetails: file`: the daemon writes one HTML page per turn to `toolDetailsDir/<platform>/<session>/<turn>.html` (default dir `~/.claude-threads/tool-details`) plus an `index.html` per session, and with `toolDetailsUrl` the summary line links to the page from the first tool on.

```yaml
toolActivity: summary
toolDetails: file
toolDetailsUrl: https://agents.example.com/tool-details   # optional; the summary line links here
# toolDetailsDir: /srv/tool-details                       # optional; default ~/.claude-threads/tool-details
```

- The daemon only writes. Serving the directory is the operator's job, and **it must sit behind auth**: the pages hold command lines and outputs. Directories are 0700, files 0600. (We serve ours from the host with a two-line unit behind Caddy basic auth.)
- ANSI CSI and OSC sequences are stripped, HTML is escaped, path segments are injectively encoded (a session id of `..` cannot escape, `a:b` and `a/b` cannot collide).
- Pages are captured at queue time, so a reset or a new turn never rewrites a pending page. Writes are sequential per session.
- A write failure is reported once in the channel and the sink stops for the session; the reply itself is unaffected.
- `toolDetailsDir` / `toolDetailsUrl` are validated (non-empty path; http(s) URL with a host, no query or fragment) and rejected with any other `toolDetails`.

Spec: `docs/quiet-tools-spec.md` § file. 13 new tests. Reviewed by Gemini, Codex and CodeRabbit; findings folded in.


anneschuth (changes requested):
We gave this delta (the four commits on top of #534) a dedicated security review, since it writes attacker-influenced content to HTML files an operator may serve over HTTP. Headline: **no XSS or traversal path found** — the design holds up well. tsc/lint clean, 3340 tests pass, and both red-checks were genuinely red (removing `escapeHtml` and neutering `safeSegment` each failed the corresponding tests).

What we verified and confirmed:
- **HTML injection**: every dynamic interpolation is escaped in element content; the only attribute context is a numeric `href`; truncation happens upstream in #534 *before* this layer escapes, so no split-entity hazard. No bypass found.
- **ANSI stripping**: CSI + OSC (including OSC 8 hyperlinks) stripped as claimed. Some exotic sequences survive (colon-form SGR, DCS/APC, bare ESC) but all are then HTML-escaped — cosmetic garbage in a `<pre>`, never markup. Fine as-is.
- **Permissions & HOME**: recursive mkdir applies 0700 to *all* created levels (we checked intermediates, not just the leaf), files 0600, and `~` expansion happens in the bot process — correctly respecting the CLAUDE.md invariant about the account pool overriding HOME per session. Nice.
- **Queue-time capture, sequential writes, failure latch**: all behave exactly as described; the reply path can never throw from the sink.

Two changes requested:

1. **DM sessions drop the file-sink config** — `src/index.ts:753` (the DM-discovery path) still calls `resolveToolActivity` without the new `fileOptions` argument; only the main platform path was updated. A parent with `toolDetails: file` yields DM instances silently writing to the *default* dir and with no link on the summary line. Same class of bug as the one we flagged on #529 — that `deriveDmConfig` spread makes the fix one line: pass `{ dir: dmConfig.toolDetailsDir, url: dmConfig.toolDetailsUrl }` there, plus a test. (This DM-discovery call-site pattern has now bitten three times in one day; a shared helper that resolves *all* per-platform overhead in one place might be worth a follow-up.)

2. **`safeSegment`'s injectivity claim is false above 0xFF** — `charCodeAt(0).toString(16)` emits variable-width hex with no delimiter, so `safeSegment(' AC')` and `safeSegment('€')` both yield `_20AC` (verified live). Not attacker-reachable today (platform ids are operator config; Slack ts / Mattermost ids are ASCII), but the PR claims injectivity and the tests only cover ASCII. One-line fix: encode per UTF-8 byte or use fixed-width `padStart(4)` with a distinct marker for the wide form.

Informational, no action needed: `https://x?` / `https://x#` pass the URL validator (operator config, broken link at worst); turn pages aren't written atomically so an HTTP reader can glimpse a truncated page mid-write; a pre-planted symlink inside the 0700 tree would be followed, but planting one requires the daemon user's own access — standard for this sink class. Also the delta adds 10 tests, not the stated 13 — same stale-counter gremlin as your other PRs.

With those two fixes this is merge-ready once #534 lands.

---

anneschuth:
Status: **two changes, then it merges behind #534** — the DM `fileOptions` at `src/index.ts:753`, and `safeSegment`'s injectivity above 0xFF (fixed-width `padStart(4)` or per-UTF-8-byte).

No need to touch this until #534 is settled; I don't want you rebasing a stack twice. Flagging only so it doesn't read as ignored — the security pass on this one came back clean and I was impressed by it.

kaza:
Both changes in, plus three defects the review round turned up.

**1. `safeSegment` injectivity.** You were right, and my first fix was wrong in the same way. Encoding per UTF-8 byte fixed `' AC'` vs `'€'` and introduced a quieter collision one level down: `TextEncoder` maps *every* unpaired surrogate to U+FFFD, so `'\uD800'` and `'\uDC00'` both became `_EF_BF_BD` — all 2,048 of them, plus a literal U+FFFD. I had restated the injectivity claim without testing the case that breaks it, which is precisely what your review was about.

It is now `_XXXX`, four hex digits per UTF-16 code unit — the fixed-width form you offered. Lone surrogates stay distinct, and it drops the `TextEncoder` and the regex `u` flag. Tests cover the multibyte, astral and lone-surrogate cases.

One thing I did **not** fix, deliberately: ASCII case is preserved, so two ids differing only in case still share a directory on a case-insensitive filesystem. Escaping case would roughly double every segment for a collision that needs a hostile config. The doc comment no longer claims the segments "survive any filesystem" — it says what it means. That claim has been wrong twice now; I would rather narrow it than widen the encoder.

**2. DM sessions dropping the file config.** Fixed, but not as a one-liner. Both call sites now go through `resolvePlatformTools(config, fieldPath)`, which reads all four fields off the config object. Your parenthesis was the actual finding: this call-site pattern has bitten repeatedly, and an argument list is what makes the omission invisible.

My first attempt at that was theatre — the parameter was an all-optional shape, so `{}` and `{ toolActivity, toolDetails }` both type-checked and the "guard" documented an intent it did not enforce. It now takes `PlatformInstanceConfig`, so a stripped literal fails to compile; my own test literals had to be widened to prove it. Worth knowing: `src/index.ts` still has no test harness, so the call site is compiler-guarded, not test-guarded.

**Three more, from the review round:**

- **A resumed session overwrote its own history.** Every resume builds a new sink for the same session id, and `turn` started at 1 unconditionally — so the first tool clobbered `1.html`, and `turnEnded` rebuilt the index from only this process's turns, unlinking every earlier page. Since every bot restart resumes sessions, the feature was quietly losing the record it exists to keep. The counter now continues from the highest `<n>.html` on disk and the index is listed from disk.
- **`https://x/details?` and `https://x/details#` passed validation.** They parse to an empty `search`/`hash`, so the property check saw nothing wrong — and the appended path then lands inside the query or fragment, so every link resolves to the base. You had this as informational; it is one line, and a link that silently 404s seemed worth more than that. Checked on the raw string now.
- The `u` flag question resolved itself when the encoding changed.

Two regression tests for those, both verified red.

3464 tests pass; typecheck, eslint and `git diff --check` clean. Reviewed by Gemini (found the surrogate collision independently), CodeRabbit (no findings) and Codex (the three above, and confirmed the fixed-width encoding injective over every UTF-16 code unit and every astral pair).

Stacked on #534, which now has your two fixes plus two more Codex found there. Ready when that one is.

anneschuth (approved):
Approved on its own delta. It lands after #534: that branch has one regression to fix (a failed plain update to the header post duplicates the paragraph into both posts) and needs current main merged in, and this one inherits both. Once #534 is green again, rebase this on it and I will merge them in order.

Thanks for the round, and for writing down the surrogate collision your first fix introduced; that note is worth more than the fix.

Both requested changes are in, and the three defects you found on the way are genuinely fixed. I re-ran the security pass on the new head and found nothing that blocks. Details below, ranked.

**Requested changes, verified**

1. `safeSegment` injectivity: `src/operations/tool-details/file.ts:60`. I ran the encoder over every single UTF-16 code unit (0 collisions), every two-unit string over 0x00..0xFF (0 collisions), and checked that no one-unit output equals any two-unit output (0). `..`, `../..`, `..\..`, NUL, `/`, `_0041` as a literal, U+FFFD, lone `\uD800` and `\uDC00`, and an astral pair all produce distinct `[A-Za-z0-9_-]` segments that `path.resolve` keeps inside the base. The case-fold caveat is stated in the doc comment rather than claimed away; I agree with leaving it.
2. DM instances keep the file config: `src/config/types.ts:131` reads all four fields off the config object, and both call sites use it (`src/index.ts:696`, `src/index.ts:749`). The parameter is the whole `PlatformInstanceConfig`, so a stripped literal fails to compile. `src/platform/dm-discovery.test.ts` covers the derived config carrying dir and URL.

**Your three follow-up fixes, verified**

- Resume continues the numbering: `file.ts:73` reads the highest `<n>.html` once, synchronously, and `file.ts:97` starts from there; `file.ts:123` lists the index from disk. This also covers a new session started later in the same thread: same session id, same directory, old pages kept and listed. `MessageManager` is only constructed on start and resume (`src/session/lifecycle.ts:1138`, `:1562`), and `!cd` reuses it, so two sinks never race on one directory.
- Empty delimiters: `src/config/types.ts:149` checks the raw string, and the test covers `?` and `#` with nothing after them.
- The `u` flag is gone with the encoding change.

**New findings**

1. **No retention for the pages (medium, can be a follow-up issue).** Thread logs are purged after 30 days by the hourly scheduler (`src/cleanup/scheduler.ts:193`); the tool-details tree has no purge at all and no mention in the data storage table in `CLAUDE.md`. These pages are the more sensitive artifact of the two, since they are the ones an operator serves over HTTP, and they accumulate per turn for every session forever. Either hook a `retentionDays` for the tree into `CleanupScheduler` next to the log purge, or state in `docs/CONFIGURATION.md` that retention is the operator's job. I would not hold the merge for it, but it should not ship in a release without one of the two.
2. **One URL tree for every platform instance (low, docs).** `toolDetailsUrl` and `toolDetailsDir` are per platform, but the default puts every instance under the same directory, and the natural setup (one Caddy basic-auth block) makes one password cover every platform's sessions. That crosses the per-platform privacy boundary the memory layer treats as hard. Worth one sentence in the docs: if the audiences differ, give each platform its own dir and URL. Same paragraph: do not enable directory listing, since the segments are the thread ids and enumerate every session.
3. **Slack unfurls the `details` link (low).** Thread replies are posted with `unfurl_links` on (`src/platform/slack/client.ts:1003`), so Slack's crawler fetches the page URL. Behind auth this is a harmless 401. Without auth, which the docs already forbid, the preview of a page holding command lines lands in the channel and on Slack's servers. Worth noting next to the "behind auth" warning; disabling unfurls on the reply post would also disable them for legitimate links in the reply, so I would not change the code.
4. **Nit:** `src/operations/message-manager.ts:329` repeats the default path as a literal while `DEFAULT_TOOL_DETAILS_DIR` is exported for it; the fallback is also dead, since `resolveToolActivity` always sets `dir` for `file`.
5. **Nit:** the comment says 3464 tests; this head runs 3382.

**Mergeability**

The branch conflicts with current `main` in `CHANGELOG.md`, `src/index.ts` and `src/platform/dm-discovery.test.ts`. All three are textual: the changelog entries, the MCP posture changes from #563 and #565 in the platform loop, and #567's DM test sitting where your new DM test goes (keep both). The #535 delta on top of #534 merges clean, so the order stays: rebase #534 on `main`, merge it, then rebase this one.

**What I verified**

- `tsc`, `eslint`, `knip` and `git diff --check` clean; 3382 tests pass.
- HTML: every interpolation is escaped in element content, including the `<title>`, so a tool line cannot close the title or open a tag; the only attribute is the numeric `href`. ANSI CSI and OSC are stripped before escaping; DCS, APC and colon-form SGR survive as escaped text, which is cosmetic.
- Directories are created 0700 at every level, files 0600, and `~` expands with the bot's own `homedir()`, not the pooled account's `HOME`.
- Queue-time capture holds: a reset while a write is pending leaves that write on its own turn and body. A write failure posts once, latches, and the summary line drops the link on the next tool. The executor guard keeps a tool-less turn from bumping the counter on the double reset in the respawn path.
- The feature is off unless `toolDetails: file` is set; `full`/`none` stay the defaults.

anneschuth:
I pushed the fix for my #534 blocker to this branch rather than sending you round the loop again, since I had already written the failing test while checking whether the defect survived the stack. Commit `7a3365b`. Please read it as a proposal and say so if you disagree.

**The defect was still here.** `content.ts`, the catch block of `tryUpdatePost`, recorded the attempted body for every caller. Reproduced with the sequence from my review: post 1 holds `para one`, a tool ends, `para two` arrives, the plain update is genuinely rejected, and `para two` then lands in post 1 (via the header re-render) *and* post 2 (via the pending copy the failure branch left in place).

**I was wrong about one thing in that review, and it makes the fix simpler than I described.** I wrote that removing the assignment breaks the rollback test. It does, but only because the assignment was in the helper. Moving it to the split's first-part `onFailure` satisfies both cases: your rollback test splits, so it goes through exactly that path and stays green, while the plain update path no longer records an attempt whose text is still pending. Both tests pass together. The two scenarios only conflict while the policy lives in the helper.

That is also what the helper's own comment says to do:

> Do NOT bake a single "correct" state reset into this helper: the 5 original sites drifted (some clear currentPostContent, some don't), and hiding those differences would be a regression hazard.

The catch block had quietly become one of those baked-in resets.

**I checked the other five call sites** rather than assuming the split was the only one. All five leave `pendingContent` in place on failure, so recording the attempt would duplicate in each; only the split's first part drops it unconditionally (`currentPostId = null`, `currentPostContent = ''` before `createNewPost`). So the one placement is correct and the three sibling `handleSplit` early returns need nothing.

Red-green, since this is a regression fix: with the assignment moved back into the helper the new test fails and the other 36 pass; with it at the call site all 37 pass. Full suite 3531 pass / 0 fail, lint, typecheck clean.

I also merged current main into the branch (`resolveReconnectPolicy` from #531 landed next to the tool exports in `src/config/index.ts`).

Two non-blocking things from my earlier review are unchanged and I am not asking for them here: the uncancelled 500 ms timer in `tool-details/thread.ts` that `reset()` does not clear, and the `!cd` abandoned-post behavior.

One process note: this PR carries all of #534's code, so #534 is now redundant. Once this lands I will close #534 pointing here, unless you would rather split them back apart.

anneschuth:
Reviewed properly now, including the fix I pushed yesterday. The `toolActivity` half is good work and I want it. My objection is to `toolDetails: file`, and it is a product objection rather than a code one.

**First, the fix I pushed (`7a3365b`) held up under challenge.** I had it audited against all six `tryUpdatePost` call sites, and the placement is right for a reason worth stating: the discriminator is whether the failure branch drops the pending content. Only the split's first part does (`currentPostId = null; currentPostContent = ''` before `createNewPost`, with no `clearFlushedContent` on failure), so only there is the header re-render the last chance to restore it. The other five leave `pendingContent` in place, so recording the attempt would duplicate. Both directions are now pinned: putting the assignment back in the helper fails the new repro, removing it entirely fails your rollback test. 37/37.

**Now the part I would change: `file` writes an unbounded, unswept pile of plaintext tool output.**

`src/cleanup/scheduler.ts` sweeps exactly two things — `cleanupOldLogs(this.logRetentionDays)` and `cleanupOrphanedWorktrees()`. Nothing references `tool-details`. Meanwhile this repo's own retention table gives thread logs 30 days and orphaned worktrees 24 hours, and `~/.claude-threads/tool-details/` would sit directly beside `logs/` and `worktrees/`, growing forever. A 40-tool turn is 41 writes of a page holding full command lines and their output; a resumed session appends indefinitely; sessions are never pruned. There is not even a command to list or clear it, where `!memory` at least has `forget`.

That is the single most load-bearing convention the project documents about itself, and this is the one store that ignores it.

**The security shape is handled honestly but fails open.** Credit where it is due: 0700 dirs, 0600 files, HTML escaped, ANSI stripped, `safeSegment` injective at fixed width with the surrogate collision genuinely fixed and the case-insensitivity limit stated rather than argued away. Nothing is served by the daemon, which is the right call. But tool output contains secrets, file contents and command output, and `toolDetailsUrl` exists specifically to put a clickable link to it in a chat channel. Whether the reader can open it depends entirely on operator-configured auth this project cannot verify. Compare #575, which exists precisely because data leaving the operator's infrastructure needed one switch, and which **fails closed** on a malformed value. This fails open into a directory nothing ever sweeps.

**And the requester does not want it.** @thejdubb02 said on #505, plainly, that he would run `toolActivity: hidden` and that `toolDetails: thread` being native on both platforms with no infrastructure is the right default. `thread` delivers the same content with no retention policy, no auth, no disk growth, and access control inherited from the channel the reader is already in. `file` is your deployment's answer — a legitimate one, you serve it behind Caddy with basic auth — but it becomes permanent public config surface with a documented external dependency the project can neither ship nor verify.

**So: I would like to take `toolActivity` now and leave `file` out.**

Concretely, my preference:

1. Drop `toolDetails: file`, `toolDetailsDir` and `toolDetailsUrl` from this PR. That leaves #534's content, which I have already reviewed to the bottom twice, plus the header fix.
2. Move the CHANGELOG entries to `## [Unreleased]` — 1.36.0 and 1.36.1 were cut from under this branch, so they currently sit inside released sections.
3. Re-land `file` as its own PR with the retention question answered: swept by `cleanup/scheduler.ts` next to `cleanupOldLogs`, a configurable `retentionDays`, and a row in the CLAUDE.md retention table.

If you would rather keep `file` here, then (1) becomes: add it to the cleanup scheduler with a retention default matching thread logs, document it in the retention table, and de-duplicate the default-dir literal (`DEFAULT_TOOL_DETAILS_DIR` in `types.ts` is re-typed as a string at `message-manager.ts:339`; one of them will drift). I would take that too — it is your call which shape, not mine, and I am not going to hold the `toolActivity` work hostage over it.

**One non-blocking thing.** `ToolActivityExecutor.reset()` returns early when `stats.started === 0`, which skips the sink reset and its 500 ms timer. I had it probed: not production-reachable, because every real `sink.append` goes through `tae.execute()`, which increments `started` first. So it is a latent invariant rather than a bug, but it is an invariant held two files away from the code that depends on it. Worth making the reset unconditional.

Tell me which shape you want and I will help get it there. And to say it plainly: finding your own surrogate collision, calling your first guard theatre when it type-checked but did not enforce, and narrowing a doc claim rather than widening the encoder — that is better than most review responses I get.

Issue #388, opened 2026-05-21
Allow any channel user to interact


jonaaix:
**Problem**
It’s a bit inflexible that I have to define the allowed user names in the config. This means any user who joins the channel spontaneously can’t interact with the bot.

**Proposed solution**
Add option to allow all users in a channel to interact with the bot.

**Sidequest**

Yesterday I set up Slack for the first time, and I can (most of the time) interact with the bot, while the lib reports that I am not authorized. Looks like a bug

<img width="545" height="227" alt="Image" src="https://github.com/user-attachments/assets/7120d3c0-ebfa-45b8-b65d-87ee0065c174" />

roskee:
This is already possible. leave the allowed users list as empty and everyone in the channel will be allowed to interact with claude

anneschuth:
On the main request: this already works. Leave `allowedUsers` empty (or omit it) and every member of the channel can interact with the bot. The check is in `isUserAllowed` (`src/platform/base-client.ts`): an empty allowlist returns `true` for everyone. `SETUP_GUIDE.md` documents it too (`allowedUsers: []  # empty = allow everyone`). Thanks @roskee for pointing this out.

On the sidequest (the "not authorized" warning while interaction still worked): that was a real authorization bypass, not just a cosmetic warning. An unauthorized user could reach Claude through the resume path, which ran from persisted session state without an identity check. Fixed in #390, which makes authorization deny-by-default on every user-driven path that invokes Claude (`startSession`, `sendFollowUp`, `resumePausedSession`, and resume-from-reaction), all routed through one helper. Closing once that merges.

Issue #523, opened 2026-09-01
Offer: `!usage` — subscription quota windows, with `!usage all` across multiple CLAUDE_CONFIG_DIR profiles


kaza:
`!cost` answers "what did this session cost". The question that actually stops work is "how close is this seat to its limit, and which limit" — and there's no way to ask it from Slack today. On a box where several people share a bot, the person who hits the wall is usually not the person watching a terminal.

That gets worse with more than one profile. Claude Code isolates seats by `CLAUDE_CONFIG_DIR`, so anyone running work and personal seats — or a bot seat beside a human one — has several on one machine. Our agent box has four; my laptop has six. The only way to see them today is to open each one.

Concretely: a seat on our box had been unauthenticated for **two days** and nothing told us. We found it by building this.

## Proposal

- `!usage` — the quota windows for the profile this session runs under.
- `!usage all` — every Claude profile discoverable on the machine, one block each.

Same subcommand shape as `!worktree`. Default stays narrow; `all` is opt-in.

Output is the three windows Claude Code shows itself:

```
vvs (someone@example.com)
─────────────────────────
Current session
███░░░░░░░░░░░░░░░░░░░░░  11% used
Resets 11:19pm (UTC)

Current week (all models)
█░░░░░░░░░░░░░░░░░░░░░░░   2% used
Resets Sep 6 at 11:59pm (UTC)

Current week (Fable)
█░░░░░░░░░░░░░░░░░░░░░░░   1% used
Resets Sep 6 at 11:59pm (UTC)
```

## What we've already built and verified

Working on a branch, running against real seats on **macOS (Keychain, 5 profiles)** and **Linux (credentials file, 2 profiles)**. Not a sketch — but not a PR yet either, because the questions below are yours to answer first.

Things worth knowing before anyone implements this:

**⚠️ `weekly_scoped` is a separate bucket from `weekly_all`.** We measured a seat reading 62% overall that was simultaneously at 100% on its model-scoped weekly limit. Reporting only the headline number tells you that you have headroom when you have none, so the per-model row has to be visible.

**⚠️ Only `limits[]` has a dependable shape.** `GET https://api.anthropic.com/api/oauth/usage` also returns top-level `five_hour`/`seven_day` and a rotating cast of internal codename buckets (`tangelo`, `iguana_necktie`, `nimbus_quill`, …), mostly null. Reading those would rot. Both of these headers are required or it answers **429**, which reads like a rate limit and sends you looking in entirely the wrong place:

[...]

PR #538, opened 2026-09-05
fix: !stop makes a direct-channel-mode channel permanently deaf

### LINKED ISSUE #537: !stop makes a direct-channel-mode channel permanently deaf
### What happens

In direct channel mode, `!stop` makes the channel permanently unresponsive. Nothing can start a session in it again — not a message, not `!help`, not another `!stop`. Only deleting the persisted record or restarting recovers it.

Reproduced on demand, three times:

```
10:20  hi          → 👀, session starts, bot replies
10:21  !stop       → "🛑 Session cancelled by @almir"
10:21  !help       → nothing. No reply, no reaction, no log line.
10:21  hi          → 👀 … and nothing, forever
       ack this    → 👀 … and nothing
```

### Why

Two lookups disagree about what a persisted session is.

`registry.getPersistedByThreadId()` → `findByThreadIdAnyState()` returns soft-deleted records **on purpose**, so a plain reply can revive a session `cleanStale()` tombstoned at startup (there is a test for it: *"returns soft-deleted sessions too (reply-resume after restart)"*). That lookup is the gate into the paused-session branch in `message-handler.ts`.

`resumePausedSession()` resolves its state through `sessionStore.load()`, which **skips** records with `cleanedAt`.

So for a soft-deleted record:

| | says |
|---|---|
| the gate (`findByThreadIdAnyState`) | "a paused session lives here" → take the paused branch |
| the sink (`load()`) | "No persisted session found" → `return`, silently |

The paused branch has already claimed the message, so the new-session path never runs either. The thread is unreachable in both directions.

`!stop` on a paused thread calls `cancelPausedSession()` → `softDelete()`, which is what sets `cleanedAt`. In a normal thread you'd never notice: you'd open a new thread. In DCM **the channel is the session**, so there is no new thread to open — the channel is simply dead.

Worth noting the same failure mode is already documented in `session-store.ts`, in the DCM guard added to `cleanStale()`:

> *"Tombstoning one after a quiet hour makes the whole channel permanently unresponsive (every message dies on 'No persisted session found')."*

That guard closed one route in. `cancelPausedSession()` is a second one, and by the same reasoning the "reply-resume after restart" guarantee the any-state lookup exists for has never actually worked — the sink drops those records too.

### Second, smaller problem

The paused branch consumes **every** command except `!stop` in silence — no reply, no log:

```ts
const pausedParsed = parseCommand(content);
if (pausedParsed) {
  if (pausedParsed.command === 'stop') { /* … */ }
  // All commands in paused state are consumed (not passed as prompts)
  return;
}
```

`!help` needs no session at all, and it is the first thing anyone tries when a thread stops answering. So the one command that could have explained the situation was also the one guaranteed to say nothing — which makes a stuck thread look exactly like a bot that has gone deaf at the socket. That misdirection is most of why this took hours to find instead of one message.

### Environment

- claude-threads v1.32.1, Claude Code 2.1.258
- Slack, direct channel mode, dynamic channels (channel-as-task)
- Linux, bun

### Fix

PR #538 — the resume sink uses the same any-state lookup the gate uses and clears the tombstone it revives (after the #388 authorization gate, so a refused resume cannot launder a soft-deleted session back into the visible set); session-less commands answer in a paused thread instead of vanishing, and drops that remain are logged. Five regression tests, each verified red against the code before it.

kaza:
Fixes #537.

In direct channel mode, `!stop` made the channel permanently deaf. Nothing could start a session in it again — not a message, not `!help`, not another `!stop`.

### The bug

Two lookups disagreed about what a persisted session is.

`registry.getPersistedByThreadId()` → `findByThreadIdAnyState()` returns soft-deleted records **on purpose**, so a plain reply can revive a session `cleanStale()` tombstoned at startup (there's a test for it). That lookup is the gate into the paused-session branch.

`resumePausedSession()` resolved its state through `sessionStore.load()`, which **skips** records with `cleanedAt`.

So a soft-deleted record was claimed by the paused branch — which meant the new-session path never ran — and then dropped at the sink on `No persisted session found`. Unreachable in both directions.

`!stop` on a paused thread calls `cancelPausedSession()` → `softDelete()`, which is exactly what sets `cleanedAt`. In a normal thread you'd shrug and open a new one. In DCM **the channel is the session**, so there is no new thread to open.

The same failure mode is already documented in `session-store.ts`, in the DCM guard on `cleanStale()`:

> *"Tombstoning one after a quiet hour makes the whole channel permanently unresponsive (every message dies on 'No persisted session found')."*

That guard closed one route in. `cancelPausedSession()` is a second — and by the same reasoning, the reply-resume-after-restart guarantee the any-state lookup exists for has never actually worked, because the sink drops those records too. This PR is the fix at the sink, so both routes close at once.

### The change

**`resumePausedSession` uses the same any-state lookup the gate uses.** The local `findPersistedByThreadId` helper it used to call is removed; `findByThreadIdAnyState` carries the same platformId privacy boundary, so cross-platform scoping is unchanged (the existing collision test still covers it).

**It clears the tombstone it revives** — written back to the store, not just mutated on the object, so a restart before the next save can't put the thread back in the trap. The write-back sits *after* the #388 authorization gate: a refused resume must not launder a soft-deleted session into the visible set. There's a test for that specifically.

**Session-less commands answer in a paused thread.** The branch consumed every command except `!stop` in silence. `!help` needs no session and is the first thing anyone tries when a thread goes quiet, so the one command that could have explained the situation was guaranteed to say nothing — which makes a stuck thread look exactly like a bot gone deaf at the socket. Session-less commands now run through the executor's first-message mode; commands that genuinely need a live session are still refused, but the drop is logged rather than vanishing.

### Tests

Five regression tests, each verified red against the code before it:

- resumes a soft-deleted record instead of dropping the message
- clears the tombstone it resurrected
- writes the revived record back to the store, not just the object
- does not revive the tombstone for a refused user
- `!help` still answers in a paused thread; a session-only command is dropped but logged

`bun test src/` — 3309 pass, 0 fail. `tsc --noEmit` and eslint clean.

### How it was found

Reproduced on demand in a task channel:

```
hi        → 👀, session starts, bot replies
!stop     → "🛑 Session cancelled"
!help     → nothing. No reply, no reaction, no log line.
hi        → 👀 … and nothing, forever
```

### DIFF
--- a/src/message-handler.test.ts
+++ b/src/message-handler.test.ts
@@ -802,6 +802,86 @@ describe('handleMessage', () => {
       expect(session.cancelPausedSession).toHaveBeenCalledWith('thread1', 'test-platform');
     });
 
+    test('a stopped thread starts a FRESH session, it does not resume', async () => {
+      // The end-to-end shape of the bug, from the operator's side. After
+      // `!stop` the record is a 'stopped' tombstone, which the paused-session
+      // gate now hides — so the next message must reach the new-session path.
+      //
+      // Getting this wrong in either direction is bad: leave the record
+      // visible and the thread is trapped (the original bug); revive it and
+      // `!stop` silently undoes itself, resurrecting a conversation that has
+      // already been distilled into channel memory as ended.
+      (session.registry.getPersistedByThreadId as any).mockReturnValue(undefined);
+      (session.getPersistedSession as any).mockReturnValue(undefined);
+
+      const post: PlatformPost = {
+        id: 'post1',
+        platformId: 'test',
+        channelId: 'channel1',
+        userId: 'user1',
+        message: '@claude-bot pick this back up',
+        rootId: 'thread1',
+        createAt: Date.now(),
+      };
+      const user: PlatformUser = { id: 'user1', username: 'allowed-user', displayName: 'User' };
+
+      await handleMessage(client, session, post, user, options);
+
+      expect(session.resumePausedSession).not.toHaveBeenCalled();
+      expect(session.startSession).toHaveBeenCalled();
+    });
+
+    test('a bare !stop on a stopped thread is a no-op, not a new prompt', async () => {
+      // Once stopped tombstones route to the new-session path, `!stop` reaches
+      // it too. It is not `worksInFirstMessage`, so without the guard it falls
+      // through and starts a session whose opening prompt is the literal text
+      // "!stop" — a bot that answers a request to stop by starting.
+      (session.registry.getPersistedByThreadId as any).mockReturnValue(undefined);
+      (session.getPersistedSession as any).mockReturnValue(undefined);
+
+      const post: PlatformPost = {
+        id: 'post1',
+        platformId: 'test',
+        channelId: 'channel1',
+        userId: 'user1',
+        message: '@claude-bot !stop',
+        rootId: 'thread1',
+        createAt: Date.now(),
+      };
+      const user: PlatformUser = { id: 'user1', username: 'allowed-user', displayName: 'User' };
+
+      await handleMessage(client, session, post, user, options);
+
+      expect(session.startSession).not.toHaveBeenCalled();
+    });
+
+    test('a non-allowlisted user gets no commands in a paused thread', async () => {
+      // The paused branch must not become the one place an outsider can run
+      // first-message commands. `worksInFirstMessage` covers `!worktree list`
+      // and `!worktree switch`, which post repository branches and absolute
+      // paths, and several handlers never consult `ctx.isAllowed` themselves —
+      // so the refusal has to happen before the executor, exactly as the
+      // new-session path does it.
+      const post: PlatformPost = {
+        id: 'post1',
+        platformId: 'test',
+        channelId: 'channel1',
+        userId: 'user1',
+        message: '!help',
+        rootId: 'thread1',
+        createAt: Date.now(),
+      };
+      const outsider: PlatformUser = { id: 'u9', username: 'random-person', displayName: 'Nope' };
+
+      await handleMessage(client, session, post, outsider, options);
+
+      // `!help` is the mildest thing behind that gate and the easiest to
+      // observe; the same refusal is what keeps `!worktree list` and
+      // `!worktree switch` — which post branch names and absolute paths —
+      // from answering an outsider here.
+      expect(client.createPost).not.toHaveBeenCalled();
+    });
+
     test('!help still answers in a paused thread', async () => {
       // Regression: every command except !stop was consumed here in silence.
       // !help needs no session at all, and it is the first thing anyone tries
--- a/src/message-handler.test.ts
+++ b/src/message-handler.test.ts
@@ -906,6 +906,54 @@ describe('handleMessage', () => {
       expect(helpText).toContain('!stop');
     });
 
+    test('!worktree remove is dropped, not silently reported as done', async () => {
+      // `worksInFirstMessage` is not "needs no session". !worktree remove
+      // carries that flag and still calls active-session-only methods: routed
+      // through the first-message executor in a paused thread it finds no
+      // session, does nothing, and returns handled — succeeding at nothing,
+      // quietly, which is the exact shape of failure this branch exists to
+      // stop producing.
+      const post: PlatformPost = {
+        id: 'post1',
+        platformId: 'test',
+        channelId: 'channel1',
+        userId: 'user1',
+        message: '!worktree remove some-branch',
+        rootId: 'thread1',
+        createAt: Date.now(),
+      };
+      const user: PlatformUser = { id: 'user1', username: 'allowed-user', displayName: 'User' };
+
+      await handleMessage(client, session, post, user, options);
+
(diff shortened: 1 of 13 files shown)


kaza:
Flagging where this sits relative to #499 before you find it in the diff.

In #499 you wrote, choosing between two fixes: *"It's simpler than teaching the resume path to revive tombstones."* This PR does some of that second thing, so it deserves an explicit look rather than a quiet landing.

**What does not conflict** — most of it. The bug here is a different trigger: `!stop` tombstones through `cancelPausedSession()` → `softDelete()`, which #530's `cleanStale` exemption does not cover. The fix for that is the `EndReason` split plus hiding `stopped` tombstones from the paused-session gate, so the next message falls through to a **fresh session**. That is your instinct from #499 — a tombstoned record should not trap the thread — applied to the path #530 leaves open.

**What does conflict** — one half. `resumePausedSession` now uses the any-state lookup and revives `stale` tombstones. That is the "teach the resume path" design you passed on.

My reasoning for including it: `registry.getPersistedByThreadId()` already returns soft-deleted records deliberately (there is a test named for it — *"returns soft-deleted sessions too (reply-resume after restart)"*), and that gate **claims the message**. So a stale tombstone was routed into the paused branch and then dropped at the sink, meaning the reply-resume-after-restart guarantee that lookup exists for has never actually worked. The two lookups disagreeing is what makes a thread unreachable in both directions at once.

**Happy to cut it to your preference** — the alternative is to hide *every* tombstone from the gate, drop the sink change entirely, and let a stale record start a fresh session too. Smaller diff, resume path stays dumb, and the `!stop` bug is still fixed. The cost is the "send a new message to continue" promise on the timeout post, and deleting that named test.

Say which and I will reshape it — this is your call, not mine.

---

Also in here, both found by review rather than by me:

- **The reaction door.** `tryResumeFromReaction` finds its record by post id, so it never passes the gate. A stopped session keeps its `sessionStartPostId`, so its old posts still carry a live-looking 🔄 — `!stop` could be undone by reacting to any message from before it. Fixed at both doors.
- **`worksInFirstMessage` is not "needs no session".** `!worktree remove` / `cleanup` / `off` carry that flag and call active-session-only methods; dispatched in a paused thread they found nothing, did nothing, and returned `handled: true`. Replaced with an explicit list of the three commands that answer from something other than a session.

anneschuth (approved):
Verified in depth — this is excellent work, and the answer to the question you left us is: **keep the full shape, exactly as built.** The `EndReason` split is the right semantic line: `!stop` means the human ended the conversation (fresh session next, never a resurrection), while a stale sweep is the system's bookkeeping (revive, honoring the "send a new message to continue" promise on the timeout post). Your reasoning for teaching the sink the any-state lookup is correct — the reply-resume-after-restart guarantee that lookup exists for never actually worked, and we confirmed the old behavior was "deaf until the 3-day purge" for *any* tombstoned thread, not just DCM. The smaller alternative would have broken a promise the bot literally posts; no thanks needed for declining it.

What we verified independently (worktree of 79f66fb):
- **Authorization ordering** — the critical property. Lookup → `isRevivable` → platform resolution → the #388 gate → only then tombstone-clear + store write. We red-checked it by moving the revive block above the gate: exactly the refused-user test failed. A refused user cannot clear a tombstone, cause a store write, or learn the session exists.
- **Privacy boundary** — the sink passes `platformId` into the any-state lookup, and the delivery-side switch to composite-key lookup quietly fixes a latent first-match-across-platforms hole. Nice catch beyond the description.
- **Purge semantics** — revival deletes `cleanedAt` and hits disk before `resumeSession` runs, so a revived record can't be purged; still-stopped tombstones purge normally; a failed resume after write-back is retried, not a trap.
- **Reaction door** — red-checked: removing the stopped-tombstone check fails exactly the 🔄-resurrection test.
- Four independent red-checks all genuinely red; tsc/lint clean; **3318 pass / +16 tests** (your body says 3309/+5 — undercounting yourself this time, which is a first for the counter).
- Clean textual merge with #530 in either order.

Merging now. One small cleanup PR would tie off four leftovers, none blocking:
1. `PAUSED_SAFE_COMMANDS` lists `'usage'`, but no such command exists in the registry — `!usage` in a paused thread falls to the misleading "needs an active session" debug line. Remove the entry (or wire it up when #523's work lands).
2. The first ⚠️ paragraph in `mock-session-context.ts`'s comment says deriving the any-state scan from `load()` is "the one thing this mock must not do" — and the code then correctly does exactly that. Leftover from an earlier revision; delete the paragraph.
3. With this merged, #530's `cleanStale` DCM-skip comment ("channel goes permanently deaf") is no longer literally true — reword it to the lifecycle-policy rationale (DCM sessions don't idle out; the channel owns their lifetime).
4. The `isDcmThreadId` swap we asked to carry over from the #530 review (`session-store.ts`, the `startsWith('dcm:')` at the cleanStale skip).

One thought for that PR or later: a second `!stop` (or `!escape`) in a session-less thread is now a silent debug-only no-op where it used to answer — defensible for loop safety, but a rate-limited one-liner for authorized users would be kinder. Your call.

---

PR #540, opened 2026-09-05
Clear CLAUDE_CONFIG_DIR when an account is selected by HOME (fixes #539)

### LINKED ISSUE #539: Pooled accounts can all run on the bot's own seat: CLAUDE_CONFIG_DIR outranks HOME
### What happens

With a `claudeAccounts` pool configured, **every pooled account can spawn against the bot's own seat** while being labelled with the pooled account's id.

The pool appears to work. Sessions are distributed, the sticky shows a range, `refreshAccountUsage()` returns numbers. But one account is doing all the work and paying for all of it, and the usage the router balances on is that same account's windows read N times under N different names.

Nothing errors. Nothing logs. The only symptom is a billing statement.

### Why

`buildClaudeChildEnv` (`src/claude/cli.ts`) overrides `HOME` for an account selected by `home`, and deliberately clears the two env vars that would otherwise beat the file-based credentials it is pointing at:

```ts
if (account?.home) {
  env.HOME = account.home;
  env.USERPROFILE = account.home;
  delete env.ANTHROPIC_API_KEY;
  delete env.CLAUDE_CODE_OAUTH_TOKEN;
}
```

`CLAUDE_CONFIG_DIR` belongs in that list and is missing — and it is the sharpest of the three, because **it outranks `HOME`**. Claude Code resolves its profile from `CLAUDE_CONFIG_DIR` when set, and only falls back to `$HOME/.claude` when it isn't.

### When it bites

Only when `CLAUDE_CONFIG_DIR` is set in the daemon's own environment — which is the normal way to run the bot beside a human seat on the same machine, so that the two don't share a profile. Ours is set that way:

```
$ tr '\0' '\n' < /proc/$(systemctl show claude-threads -p MainPID --value)/environ | grep -E '^HOME|^CLAUDE_CONFIG_DIR'
HOME=/home/herder
CLAUDE_CONFIG_DIR=/home/herder/.claude-vvs
```

So any operator who isolated the bot's profile *and* configured a pool has a pool of one seat wearing several names. We haven't been bitten because we run no pool yet — this was found by review while reshaping `!usage` onto `usage-probe.ts` (#523), where it would have made every reported row a copy of the bot's own quota.

An API-key account is unaffected: it is selected by `ANTHROPIC_API_KEY` and has no config dir of its own to point at.

### Fix

PR #540 — one line, `delete env.CLAUDE_CONFIG_DIR` alongside the other two, plus three tests covering the pooled, single-account and API-key cases. The first is verified red against current `main`.

### Environment

claude-threads v1.32.1, Claude Code 2.1.258, Linux, bun.

kaza:
Fixes #539.

`buildClaudeChildEnv` overrides `HOME` for an account selected by `home`, and clears the two env vars that would otherwise beat the file-based credentials it is pointing at — the inherited API key and OAuth token. `CLAUDE_CONFIG_DIR` belongs in that list and was missing, and it is the sharpest of the three because **it outranks `HOME`**.

A daemon started under its own profile has it set, which is the normal way to run the bot beside a human seat without the two sharing a profile. Every pooled account then spawns against the *bot's* config dir while carrying the pooled account's id: every session billed to one account, and `refreshAccountUsage()` reading that same account's windows once per pool entry — so the router balances a pool of one seat wearing several names. Nothing errors and nothing logs; the symptom is a billing statement.

One line, alongside the other two.

An API-key account is unaffected — it is selected by `ANTHROPIC_API_KEY` and has no config dir of its own to point at. Single-account mode is unaffected — the bot's own profile *is* the seat, so clearing the config dir there would send the session to the wrong one. Both have a test.

Three tests; the first is verified red against current `main`.

Found by review while reshaping `!usage` onto `usage-probe.ts` for #523, where it would have made every reported row a copy of the bot's own quota. Kept separate from that work because this one reaches session spawning, not just the probe, and can land on its own.

### DIFF
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -554,6 +554,14 @@ claudeAccounts:
 | `apiKey` | One of | Anthropic API key. Billed against that key; session history stays under the bot's default `HOME`. |
 | `displayName` | No | Human-readable label in UI (defaults to `id`) |
 
+⚠️ An account with `home` owns the **whole** Claude profile, not just its
+credentials: user settings, hooks and global MCP configuration are read from
+that home too. Anything the daemon inherited that could point elsewhere — an
+API key, a bearer token, `CLAUDE_CONFIG_DIR` — is cleared for that child, or
+the account you selected would be silently overridden by the one the daemon
+runs as. Daemon-level hooks and settings therefore do **not** carry into a
+pooled account; put them in the account's own home.
+
 Exactly one of `home` or `apiKey` should be set per account. Persisted sessions record which account they ran under and resume on the same one.
 
 ## Environment Variables
--- a/src/claude/cli.test.ts
+++ b/src/claude/cli.test.ts
@@ -943,6 +943,54 @@ describe('buildClaudeChildEnv: an account selected by HOME owns the whole select
     expect(env.CLAUDE_CONFIG_DIR).toBeUndefined();
   });
 
+  it('clears every inherited credential and location override, not just some', () => {
+    // Measured against the shipped CLI: ANTHROPIC_AUTH_TOKEN authenticates on
+    // its own, and CLAUDE_SECURESTORAGE_CONFIG_DIR relocates stored
+    // credentials the same way CLAUDE_CONFIG_DIR relocates the profile.
+    // Clearing a subset is the same bug with a smaller blast radius: the
+    // account we asked for, silently overridden by one we inherited.
+    const env = buildClaudeChildEnv(
+      {
+        HOME: '/home/bot',
+        CLAUDE_CONFIG_DIR: '/home/bot/.claude-vvs',
+        CLAUDE_SECURESTORAGE_CONFIG_DIR: '/home/bot/.claude-vvs',
+        ANTHROPIC_API_KEY: 'sk-inherited',
+        ANTHROPIC_AUTH_TOKEN: 'inherited-bearer',
+        CLAUDE_CODE_OAUTH_TOKEN: 'inherited-oauth',
+      },
+      { id: 'pooled', home: '/home/bot/accounts/primary' }
+    );
+
+    expect(env.ANTHROPIC_API_KEY).toBeUndefined();
+    expect(env.ANTHROPIC_AUTH_TOKEN).toBeUndefined();
+    expect(env.CLAUDE_CODE_OAUTH_TOKEN).toBeUndefined();
+    expect(env.CLAUDE_CONFIG_DIR).toBeUndefined();
+    expect(env.CLAUDE_SECURESTORAGE_CONFIG_DIR).toBeUndefined();
+  });
+
+  it('does not let an inherited bearer token outrank the API key it was given', () => {
+    // ANTHROPIC_AUTH_TOKEN authenticates by itself and wins over a key set
+    // beside it, so an API-key account would have been billed to whatever the
+    // daemon inherited.
+    const env = buildClaudeChildEnv(
+      { HOME: '/home/bot', ANTHROPIC_AUTH_TOKEN: 'inherited-bearer' },
+      { id: 'billed', apiKey: 'sk-ant-test' }
+    );
+
+    expect(env.ANTHROPIC_API_KEY).toBe('sk-ant-test');
+    expect(env.ANTHROPIC_AUTH_TOKEN).toBeUndefined();
+  });
+
+  it('leaves the parent environment untouched', () => {
+    // buildClaudeChildEnv copies before deleting; a delete that reached the
+    // parent would log the DAEMON out.
+    const parent = { HOME: '/home/bot', CLAUDE_CONFIG_DIR: '/home/bot/.claude-vvs' };
+    buildClaudeChildEnv(parent, { id: 'pooled', home: '/home/bot/accounts/primary' });
+
+    expect(parent.HOME).toBe('/home/bot');
+    expect(parent.CLAUDE_CONFIG_DIR).toBe('/home/bot/.claude-vvs');
+  });
+
   it('leaves CLAUDE_CONFIG_DIR alone when no account overrides HOME', () => {
     // Single-account mode: the bot's own profile IS the seat, so clearing the
     // config dir would send the session to the wrong one.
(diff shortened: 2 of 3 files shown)


kaza:
Review note, and one open question I would rather hand you than decide alone.

Gemini's review raised that `CLAUDE_CONFIG_DIR` may not be the only variable in this class: `XDG_CONFIG_HOME`, `XDG_DATA_HOME`, `XDG_STATE_HOME`, `XDG_CACHE_HOME` and `APPDATA` can also participate in path resolution ahead of `HOME`.

That is not hypothetical — the shipped binary references all of them:

```
$ B=$(readlink -f $(which claude))
$ for v in XDG_CONFIG_HOME XDG_DATA_HOME XDG_STATE_HOME XDG_CACHE_HOME APPDATA; do
    printf "%-18s %s\n" "$v" "$(strings "$B" | grep -c "$v")"; done
XDG_CONFIG_HOME    30
XDG_DATA_HOME      11
XDG_STATE_HOME      8
XDG_CACHE_HOME      9
APPDATA            25
```

What that count does **not** establish is precedence, or that any of them beats `HOME` for the credential/profile lookup specifically rather than for some bundled dependency's cache. I did not want to widen the deletion list on a string count.

So this PR stays at `CLAUDE_CONFIG_DIR`, which is the one I could demonstrate: it is documented as the profile override, it is the variable an operator sets to isolate a bot from a human seat on the same box, and the failure it causes is silent and expensive.

If you would like the XDG set cleared too when `home` is set, I am happy to add it here — the argument for it is symmetry: once we have overridden `HOME` for account isolation, anything inherited that can redirect path resolution defeats that isolation the same way. The argument against is that clearing `XDG_CONFIG_HOME` on a Linux host where the operator deliberately set it is a change with a blast radius beyond the pool.

For what it is worth, our own daemon carries none of them (`0` XDG or APPDATA vars in its environment), so I have no live case either way to test against.

CodeRabbit: 0 findings. Gemini: the above, plus it agreed keeping the variable for API-key and single-account mode is correct. Codex still running; I will add anything it turns up.

kaza:
Codex went further than the string counts and measured the shipped CLI directly. The isolation was narrower than this PR originally assumed, so the fix has grown — still small, but no longer one line.

**Now also cleared**, because they defeat the selection in exactly the same way the API key and OAuth token already did:

- `ANTHROPIC_AUTH_TOKEN` — authenticates on its own, and takes precedence over an API key set beside it. It was inherited by **both** branches: a pooled account could be overridden by it, and an API-key account could be billed to it rather than to the key it was handed.
- `CLAUDE_SECURESTORAGE_CONFIG_DIR` — relocates stored credentials the way `CLAUDE_CONFIG_DIR` relocates the profile.

**Deliberately not cleared, and this is the question for you.** Codex also identified the third-party provider selectors (`CLAUDE_CODE_USE_BEDROCK` / `_VERTEX` / `_FOUNDRY`) and the federation pair (`ANTHROPIC_FEDERATION_RULE_ID` / `ANTHROPIC_ORGANIZATION_ID`), all confirmed present in 2.1.261.

Those are a different category. They select a **provider**, not an account — clearing them changes which service the bot talks to, not which seat it bills. An operator running the whole daemon on Bedrock and clearing that flag per-account would find pooled sessions silently going somewhere else. The counter-argument is that `claudeAccounts` with `home` implies subscription seats, which is not a thing on Bedrock, so the combination may be incoherent anyway. I did not want to settle that on your behalf.

Same for the earlier XDG question. Both are yours; say the word and either goes in.

On the Gemini/XDG point: I checked precedence rather than presence, and could not establish that `XDG_CONFIG_HOME` beats `HOME` for the credential lookup specifically — the references may belong to a bundled dependency. So it stays out for now.

Also documented, since it was true before this PR and never written down: an account with `home` owns the **whole** profile — settings, hooks and global MCP config, not just credentials — so daemon-level hooks do not carry into a pooled account.

Four more tests, including one pinning that the parent environment is left untouched. These are deletes on a copy, and a delete that reached the parent would log the daemon itself out.

Round 2: CodeRabbit 0 findings, Gemini 1 (the XDG question above), Codex 2 (both addressed or raised here).

anneschuth (approved):
Verified independently and approving. What we confirmed:

- **The premise holds on documentation, not just measurement**: `CLAUDE_CONFIG_DIR` is the documented profile override — the CLI falls back to `$HOME/.claude` only when it's unset — so overriding `HOME` while it survives does nothing for the profile lookup. #539's live daemon environment is exactly the biting case.
- **Both consumers are fixed at once**: `usage-probe.ts` builds its spawn env through the same `buildClaudeChildEnv`, so usage balancing works under `CLAUDE_CONFIG_DIR` daemons after this — which also un-blocks the `!usage` work in #523. We checked the remaining spawn sites: `quick-query.ts` deliberately runs haiku one-shots on the bot's own seat (correct, unchanged); the memory/settings machinery is path-based with no config-dir interaction. No remaining hole found.
- **Branch scoping traced in code**: deletions only in the `home` branch (plus the bearer-token pair in the API-key branch, which is right — an inherited `ANTHROPIC_AUTH_TOKEN` outranking the handed key is the same leak); single-account mode untouched; all deletions on a spread copy with a test pinning the parent env.
- **Red-checks**: with the fix reverted, the 3 leak tests fail and the 3 behavior-preservation pins correctly pass — exactly the right split. tsc/lint clean, 3308/3308, clean merge onto main.

Answering the two questions you explicitly left to us:

1. **XDG set: keep it out**, for your own stated reason — you checked precedence rather than presence and couldn't demonstrate any of them beats `HOME` for the credential lookup specifically. A deletion list should carry only variables with demonstrated effect; string counts in the binary don't clear that bar, and clearing `XDG_CONFIG_HOME` on a host where the operator set it deliberately has blast radius beyond the pool. If a live case ever shows an XDG variable defeating seat isolation, it earns its line then.
2. **Provider selectors and federation vars: leave them inherited**, as you did. They select a provider, not a seat, and clearing them would silently reroute a Bedrock/Vertex daemon's pooled sessions. Your observation that `home` pools are subscription-shaped and the combination is likely incoherent is right — that's a docs sentence ("account pools assume first-party subscription seats; provider-selector deployments should use single-account mode"), not a deletion. Feel free to add it here or leave it to us.

The review-round transparency in this thread (what each bot found, what you took, what you declined and why) is a model for how these PRs should run. Merging.

---

PR #544, opened 2026-09-05
!usage — subscription quota per seat, over the account pool (closes #523)

### LINKED ISSUE #523: Offer: `!usage` — subscription quota windows, with `!usage all` across multiple CLAUDE_CONFIG_DIR profiles

kaza:
Closes #523.

Built to the shape you described there, which is not the shape I had. The first version read the OAuth usage endpoint directly and owned a credential layer to do it; you pointed at `usage-probe.ts` already doing the job. You were right, and the rewrite is smaller than the thing it replaced.

### What it does

`!usage` reports the subscription windows for the seat the thread runs on. `!usage all` reports every account in the `claudeAccounts` pool.

```
vvs (Max 20×)
──────────────
Current session
███░░░░░░░░░░░░░░░░░░░░░  14% used
Resets 11:50pm

Current week (all models)
█░░░░░░░░░░░░░░░░░░░░░░░   3% used
Resets Sep 2 at 3am

Current week (scoped)
███████████████░░░░░░░░░  62% used
```

Three windows, because the scoped weekly limit is a **separate bucket** from the all-models one. A seat can read 3% overall while sitting at 62% model-scoped — reporting only the headline tells you that you have headroom when you have none. `weekPerModelPct` already existed for routing; this surfaces it.

### Following your four points

1. **In scope, as its own command.** `!cost` is "what did this session spend"; `!usage` is "how close is this seat to the wall".
2. **Numbers from `usage-probe.ts`.** Nothing here touches a credential. Token refresh, refresh-token rotation and platform storage stay inside the CLI. That deleted 615 lines including every Keychain path — and with them a token that leaked into `execFile` argv on error, and a write that used the wrong Keychain account and would have made a macOS seat unreadable until someone logged in by hand. Both were real, both were found by review, and both were the price of owning that layer rather than borrowing it.
3. **Discovery is the pool.** No `~/.claude*` scan. Without a pool it reports the account the process already runs as — one row, not a row per human seat on the box.
4. **Emails behind a flag, off by default.** `usage.showEmails`.

The plan badge (`Max 20×`) is shown either way — it explains why one seat's week is four times another's and identifies nobody. Both it and the address come from `.claude.json`: measured on real seats, `organizationRateLimitTier` carries the multiplier and `userRateLimitTier` overrides it per seat. So point 2 holds for identity too, not just for the quota.

### Two things that are less than the first version had

- **`resetsAt` is the probe's phrase, not a Date.** The CLI prints "Nov 5 at 3pm" and no machine timestamp. Parsing prose into a Date only to format it back would invent a precision, and a time zone, the source never had. The line is omitted when there is no hint.
- **The scoped row names no model and carries no reset.** `parseUsageOutput` keeps the highest per-model percentage and discards which model produced it, and reports one weekly reset that belongs to the all-models window. So the row is headed "Current week (scoped)" and has no reset line. Both would be guesses. If you would like the probe to retain the winning model and its own hint, that is a small change to `parseUsageOutput` and I am happy to make it — I did not want to alter a parser the router depends on as a side effect of a display feature.

### One thing worth flagging

`parseUsageOutput` accepts a report containing **either** top-level line and defaults the missing one to `0`. So a response with only a session line renders a measured-looking weekly `0%`. That predates this PR and the router has the same exposure, but `!usage` is where a human reads the number. Making the two percentages nullable would fix it and would ripple into `usageLoadScore`, so I have left it alone and am raising it rather than deciding it.

### Reviewed

Three reviewers, two rounds, fourteen findings, eleven real. The ones that changed the code: an unscoped persisted-session lookup that could bind a row to another platform's account; the scoped row borrowing the all-models reset; a 30s-per-seat probe timeout bounded to 10s to match the router; and `plan.ts` still documenting `.credentials.json` as its source in the file whose point is that nothing reads credentials.

A fifth finding became #539/#540, merged separately — it reached session spawning, not just the probe.

`bun test src/` 3373 pass, `tsc --noEmit` and eslint clean.


anneschuth (changes requested):
Thank you — and the opening paragraph of your PR body is the best kind of contributor note: the rewrite really is smaller than what it replaced, and deleting the credential layer deleted its bugs with it. We verified independently on a worktree of 3dbb81e:

**Confirmed clean:** the no-credential claim holds (every `.credentials.json`/Keychain mention in the diff is a comment; the only file read is `.claude.json`, defensively, per-account); the `.claude.json` read resolves against the *pooled account's* home with the #540-fixed child env, so no config-dir-class bug; `usage-probe.ts`/`account-pool.ts`/`cli.ts` are provably untouched (empty diff), so the router is safe; the scoped row borrows no reset and invents no model; probe timeout genuinely 10s; `showEmails` genuinely default-off with the email absent, not masked; 3373/3373 tests, 5/5 red-checks red, clean merge on main tip. Also: your `!usage` entry in #538's `PAUSED_SAFE_COMMANDS` is now live and behaves sanely in paused threads (stale account id gets the explicit "no longer configured" row) — nice convergence.

**One change before merging:**

`handleUsage` never consults `ctx.isAllowed`. The first-message and paused paths gate on the platform allowlist before dispatch, but the **active-session path dispatches regardless** — so any channel member replying `!usage all` inside someone's session thread triggers up to N sequential `claude` spawns (10s cap each) and gets the full pool's quota and plan badges (and emails, when the flag is on). The paused branch's own comment warns that several handlers don't self-gate; this one spawns subprocesses and reveals pool internals, so it needs to. **Ask: `if (!ctx.isAllowed) return …` at the top of `handleUsage`, plus one red-checked test through the in-session path.** The other two paths pass `isAllowed: true`, so nothing else changes. (With the gate in, an audit-log line would match house style for gated commands — optional.)

**Answering what you left to us:**

- **`parseUsageOutput`'s missing-line→0**: you were right to raise rather than decide it. Our answer: yes, make the two percentages nullable — but as its own follow-up PR, because the ripple into `usageLoadScore` is router behavior and deserves its own red-green round, not a display PR's side effect. In this PR, leaving it alone is correct. (When it lands, `!usage` should render an absent line as "no data" rather than a measured-looking 0% bar.)
- **Unbound thread with a pool reports the whole pool** (plain `!usage` ≡ `!usage all` in a fresh thread): we're fine with the semantic — "the seats the router would choose between" is honest — but it diverges from the doc'd "seat the thread runs on", so add a line to `CONFIGURATION.md` and ideally a test pinning it, in this PR or the follow-up.

**Non-blocking polish, take or leave:** a short per-channel cooldown on probing (spam becomes authorized-user-only misuse once the gate is in, so this is belt-and-suspenders); the duplicated 10s timeout constant could be shared with `manager.ts`; an orphaned JSDoc paragraph above `getUsageShowEmails()` belongs to `getClaudeAccounts()`; the paused-path e2e test gap in `message-handler.test.ts`.

With the gate in, this merges — and it's a genuinely good addition, especially the three-window rendering making the scoped-bucket trap visible to the person about to hit it.

---

anneschuth:
Status: **one gate plus a test**, and the branch now needs a rebase (it's showing conflicts against main).

```ts
if (!ctx.isAllowed) return …   // top of handleUsage
```

plus one red-checked test through the in-session path. Everything else in my review was either answered or explicitly deferred — the nullable-percentages change is its own follow-up, not this PR's job.

The `CONFIGURATION.md` line pinning "plain `!usage` in an unbound thread reports the whole pool" would be good to fold in while you're rebasing.

kaza:
Gate in, plus the `CONFIGURATION.md` line.

**The gate.** `if (!ctx.isAllowed) return { handled: true }` at the top of `handleUsage`, with three tests through the real dispatch path. Worth noting how visible the hole was once tested: the red runs took ~1.1s per case, because that was the probe genuinely spawning for an unauthorized user.

**One correction, and a question it raises.** Codex traced `ctx.isAllowed` on the in-session path to `isUserAllowedInSession()` — the platform allowlist **or** a session invitee. So my first `CONFIGURATION.md` line ("restricted to the platform's `allowedUsers`") described something the code does not do: an invited collaborator outside `allowedUsers` can run `!usage all`.

I kept your literal shape and corrected the documentation instead, since an invitee is a collaborator the owner deliberately added, and "authorized in this thread" is the honest description. But the two readings genuinely differ for a pooled deployment, and it is `ctx.client.isUserAllowed(ctx.username)` if you want the pool narrower than the session — your call, one line either way. The invitee case is pinned by a test so the two cannot drift apart again.

Also correcting my own commit message: it claimed `handled: false` would forward raw `!usage` to Claude as a prompt. It would not — the caller excludes registered commands from dynamic passthrough and returns unconditionally. `handled: true` is still right (denial fully consumes the command), just not for the reason I gave.

**Both answers taken:**
- `parseUsageOutput`'s missing-line → 0 stays out of this PR. Agreed it wants its own red-green round once it touches `usageLoadScore`.
- The unbound-thread semantic is now documented in `CONFIGURATION.md` and pinned by a test.

**Two pre-existing holes Codex found in the same file**, not introduced here and not touched:
- **`!worktree list`** — the manager discards the username and the downstream handler has no authorization check, so any channel member can trigger Git subprocesses and get repository paths and branch names back.
- **Bare `!update`** — no downstream gate; triggers an npm registry check and reveals update status, versions and restart timing. `now` and `defer` are protected.

Happy to file those as an issue or fix them in a follow-up — they are the same shape as this one, and I would rather not widen this PR to reach them.

3460 tests pass; typecheck and eslint clean. Reviewed by Codex and CodeRabbit. CodeRabbit's one finding was outside this diff (the #538 tombstone flow) and I rejected it: it wanted the tombstone clear moved until after `resumeSession()` succeeds, which reintroduces the bug #538 fixed — the tombstone has to be off disk, not just off the object, or a restart before the next save re-traps the thread.

One flake to flag, unrelated to this branch: `MattermostClient > unpinPost still throws on 500` times out under full-suite load (67s) and passes in isolation. This branch does not touch `src/platform/`.

anneschuth (approved):
Thanks for the quick turnaround, and for correcting the docs instead of quietly narrowing the gate to make them true.

**The authorization question: keep `ctx.isAllowed`.**

The codebase has three tiers of command gate. Self-only (`!github-email`). Thread-authorized via `ctx.isAllowed`, which is `isUserAllowedInSession()`: `!stop`, `!escape`, `!approve` and the `!cost` / `!context` passthroughs. Owner-gated via `requireSessionOwner` in `src/operations/commands/guards.ts`: `!invite`, `!kick`, memory edits, routines and watches. An invitee already spends the seat's quota with every prompt they send and can `!stop` the session outright, so reading how close that seat is to its wall belongs in the second tier. Gating `!usage` on the platform allowlist would make it stricter than `!stop`, which is backwards. The `all` form widens the answer to the whole pool, but what it reveals is account ids the operator named, plan badges, and percentages; the address stays behind `showEmails`. That is not worth a second gate. If you ever want the pool narrower than the thread, put `ctx.client.isUserAllowed(ctx.username)` on the `all` branch only, not on the handler. Note that the paused path is already allowlist-only, because the executor passes `isAllowed: true` there only after the `client.isUserAllowed` check; that asymmetry predates this PR (`!help` has it too) and is harmless.

**Round one, item by item**

- Gate on `handleUsage`: in, `src/commands/executor.ts` top of the handler. I removed it locally and the two denial tests in `src/commands/usage-gate.test.ts` go red (2 fail, 2 pass), so they test the real dispatch path. Resolved.
- `CONFIGURATION.md` line pinning that plain `!usage` in an unbound thread reports the whole pool: in, with `accountTargets` pinned in `src/usage/accounts.test.ts`. Resolved.
- Nullable percentages: deferred by agreement. Fine.
- The non-blocking items from round one are all still open: no probe cooldown, the duplicated 10s constant in `src/usage/index.ts` and `src/session/manager.ts`, the orphaned JSDoc above `getUsageShowEmails()`, no paused-path test in `message-handler.test.ts`. None of them blocks.

**Findings**

1. `src/commands/usage-gate.test.ts:14`, the two positive-path tests spawn the real CLI. The fake pool returns `{ id: 'seat-a', home: '/nonexistent/a' }`, so an allowed user's call runs `collectUsage` into `probeAccountUsage`, which execs `claude -p /usage` with `HOME=/nonexistent/a`. On a box with a binary at the resolved path (your 1.1s per case was exactly that) a unit test runs a logged-out CLI; bun's default per-test timeout is 5s against the probe's 10s cap, so a slow CLI start fails the test. On CI and on my machine it is ENOENT and passes in 0.7s. Fix is one line: make the fake account an API-key one (`{ id: 'seat-a', apiKey: 'x' }`), which `accountTargets` turns into a note row with no IO, as `src/usage/mode.test.ts` already does. Assertions stay the same. Please fold this in while rebasing.

2. `src/usage/index.ts:139`, `collectUsage` probes the pool sequentially, so `!usage all` has a worst case of N times 10s when a seat hangs, while `probeAllAccounts` in `src/session/manager.ts` runs the same probe for the same accounts in parallel. A `Promise.all` over the targets caps the wait at one probe. The command also never feeds its results into `accountPool.setUsage`, and does not read the manager's 15s cache; both are fine, the router keeps its own view. Non-blocking.

3. `src/commands/executor.ts:104`, the "Handle !stop command" JSDoc now sits above the `handleUsage` block and `handleStop` has none. Nit.

4. `!usage foo` in a session silently reports the current seat, and on the first-message path it does not match the immediate pattern, so with a mention it starts a session with that as the prompt. Same as any other typo, mentioning it for completeness.

**Mergeability**

GitHub shows the branch as conflicting. There are exactly two conflicts, both adjacent additions: the config table row in `docs/CONFIGURATION.md` and the `usage` field next to main's new `transcription` and `mcpServers` fields in `src/config/types.ts`. I resolved both by keeping both sides in a scratch worktree; the `SessionManager` call in `src/index.ts` and `src/session/manager.ts` auto-merge, and the merged tree passes typecheck, lint and the usage suites. A mechanical rebase.

**Verified on fb3ee81**

- `bun run typecheck`, `bun run lint`, `bun run knip` clean; `bun run test` 3378 pass, 0 fail.
- The probe still goes through `buildClaudeChildEnv`: `usage-probe.ts` is untouched by the PR, so after the rebase every `!usage` spawn inherits 1.35.0's connector kill switch (`ENABLE_CLAUDEAI_MCP_SERVERS=false`) and the tuning flags for free.
- Single-account mode: `accountTargets([])` yields no targets, so one row for the process's own account, config dir honoring `CLAUDE_CONFIG_DIR` the same way the child env does. API-key accounts are listed with a note and never spawn.
- Output is one fenced code block posted through `client.createPost`, which renders as monospace on both Slack mrkdwn and Mattermost, so the bars line up on both.
- `usage.showEmails` reaches the manager through the YAML cast and defaults to false; no schema to update.
- In-session parsing: `!usage all` yields `args: 'all'`, and the paused-path allowlist from #538 already contains `usage`.

PR #545, opened 2026-09-05
Transcribe inbound audio attachments before Claude sees them (closes #519)

### LINKED ISSUE #519: Offer: transcribe inbound audio attachments (voice notes → text), provider-pluggable, ElevenLabs first
Hi Anne — another offer-before-PR, same spirit as #490.

**What we want:** send the bot a **voice note** instead of typing. Slack (and Mattermost) audio clips already arrive as ordinary attachments, and `buildMessageContent` dutifully saves them and hands Claude a path to a `.webm` — which Claude can't hear. The gap is one step: transcribe audio attachments to text before the message goes to Claude.

**Proposed shape** — provider-pluggable, so nobody is married to one vendor:

```yaml
# top-level in config.yaml
transcription:
  provider: elevenlabs        # first (and for now only) implementation
  apiKey: ...
  model: scribe_v2            # optional
  languageCode: eng           # optional; omitted = auto-detect
```

Behaviour:

- Any inbound attachment with an `audio/*` MIME type is transcribed. The prompt keeps the existing saved-file line **and** gains a `[Transcript of voice.webm:]` block, so Claude has both.
- The bot **posts the transcript back into the thread** as a quote — teammates (and the audit log) see what Claude heard, and a bad transcript is visible before Claude acts on it.
- Transcription failure is a visible warning in the thread (same channel as the skipped-files feedback); the raw file is still handed over. Nothing silent.
- No `transcription` block → behaviour unchanged. No new dependencies: `fetch` + `FormData` on Bun/Node 20.

Internally it's a ~20-line `Transcriber` interface (`transcribe(file) → text`) with an ElevenLabs Scribe implementation behind a `createTranscriber(config)` factory. Obvious later providers: OpenAI (`whisper-1`/`gpt-4o-transcribe`), local whisper, and — nice one — **Slack's own free transcript** on paid plans (it's async, so it needs a `message_changed` listener or a short poll; we skipped it for round 1).

Cost for reference: ElevenLabs Scribe is ~$0.22/hour of audio, so a voice note is fractions of a cent.

**Questions for you, before we shape the PR:**

1. Top-level `transcription:` (one vendor key for the daemon) or per-platform, mirroring `outboundFiles`?
2. Echoing the transcript into the thread — default on, or opt-in?
3. Trigger on `audio/*` only, or also `video/*` (Slack video clips carry speech too)?
4. Any objection to the provider-interface shape, or a preferred place for it (`src/transcription/`)?

We're building the ElevenLabs version on our fork this week regardless, because we need it. If you'd like it upstream, we'll cut the PR to whatever shape you answer above; if not, no hard feelings.

kaza:
Closes #519. Built to the shape you set out there.

Voice notes arrive as unreadable files today. This transcribes any `audio/*` attachment before Claude sees it, echoes the transcript into the thread, and feeds it to the watch evaluator as text.

### Your four points

1. **Top-level `transcription:`, per-platform opt-out.** The provider and its key are a property of the deployment — one vendor account per daemon — so they stay at the top level. `transcription: false` on a platform opts that channel out, gated in one place so every path to the vendor goes through it.

2. **Transcript echoed, default on.** As you put it: visible over silent. It goes into the thread as a quote, and into the prompt as a `[Transcript of voice.webm (elevenlabs):]` block.

3. **`audio/*` only.** `video/*` excluded; a `mediaTypes` knob later if it is ever wanted.

4. **`src/transcription/` with a `Transcriber` interface and a `createTranscriber(config)` factory.** Config errors fail the boot, not the first voice note.

### The two asks inside point 4, and what they turned into

**"Rebuild the provider's error text rather than echoing it."** The branch was echoing a 200-character excerpt of the raw response body into the channel. Now only the `status` CODE is repeated — an identifier from a fixed vocabulary, and only while it still matches `^[a-z][a-z0-9_]{0,48}$`, so a vendor that starts putting sentences in that field cannot publish them by surprise. The free-text `message` beside it never reaches a channel; the full body goes to the debug log, so it is withheld rather than lost.

**"The transcript should go through the same path a typed message would, not around it."** This turned out to be the substantial one, and it was a real hole rather than a polish item.

The watch evaluator runs on the one path where the bot has decided to ignore a message — which `buildMessageContent` never reaches. So a voice note arrived at the evaluator as an empty string with a file beside it: a channel watching for "deploy" fired when somebody typed it and stayed silent when somebody said it. The transcript was a courtesy for human readers and nothing more.

`transcribeForWatch` closes that. Three things guard it, because putting a *paid* call where a *free* one used to sit inherits assumptions that were only ever true for the free one:

- the platform allowlist, **before** the call — the free `evaluateWatches` beside it has always been open to anyone in the channel, but a vendor bill should not be
- whether any enabled watch actually exists — the feature flag defaults to on, so without this every ignored voice note anywhere would be uploaded to be matched against an empty list
- the downloads are removed in a `finally` — cleanup is normally tied to session teardown, and on this path no session is ever created

And when a watch does fire, the matched text now travels with it. `autoIncludeContext` pulls the thread's messages, and a voice note's platform text is empty — so without carrying it, the watch fired correctly and then asked Claude to triage an incident whose details it could not read. Quoted and fenced as data, on the same footing as the thread content beside it.

No duplicate vendor call: `src/watches/runner.ts` passes no files, so the fired session's `buildMessageContent` returns at its empty-files guard.

### One deliberate inconsistency

`resolveTranscriptionEnabled` fails **closed** on a value it cannot read, unlike `routines` / `watches` / `memory`, which fall back to enabled. Those govern whether the bot does something for you. This one governs whether a channel's audio is uploaded to a third party, and `transcription: "false"` — a quoted boolean, the likeliest way to get this wrong in YAML — must not be read as consent. Undefined still means enabled: that is the documented default and an operator who configured a provider chose it.

Happy to make it consistent instead if you would rather have one rule.

### Injection

Documented plainly, because point 4 makes it load-bearing: a transcript is user input, not narration. It reaches Claude as message text and is evaluated by watches exactly as typed text is — deliberately. So a voice note can say "ignore your instructions" as easily as a typed message can, and anyone who can post audio in the channel can try. The audio surface is exactly as trusted as the text one.

### Reviewed

Three reviewers, two rounds, eleven real findings. The ones that changed the code are above. Two rejected with reasons: removing `webm` from the generic-MIME fallback (Slack's own voice clips are `voice.webm` — the primary case, not an edge), and dropping the raw body from the debug log (the operator's own log is the only remaining copy of the diagnostic).

`bun test src/` 3362 pass, `tsc --noEmit` and eslint clean. Spec: `docs/audio-transcription-spec.md`.


anneschuth (approved):
Verified independently on a worktree of 53c4811, and this is a model implementation of the shape we agreed in #519 — approving and merging. What we confirmed:

- **The three security-sensitive requirements all hold, with genuinely red tests.** Provider errors are rebuilt, not echoed: only the HTTP status plus an allowlist-regex-validated status *code* ever reaches the channel; we sabotaged `describeErrorBody` into the naive echo implementation and five tests failed, including the hostile-flood one. Failures are visible with the raw file still handed to Claude. And the transcript goes *through* the watches evaluator, not around it — with cost gates (allowlist + enabled watches + platform flag) checked before any vendor call, which we also red-checked.
- **The API key flows only to the `xi-api-key` header** — never logged, never in an error, and transcriber exceptions can't reach the generic error post.
- **Defaults byte-identical** without the config block; zero new dependencies; per-platform opt-out fails *closed* on unreadable values (a deviation from the other flags, in the safe direction, and tested); clean merge onto current main including this morning's five merges.
- The extension fallback for generic-MIME audio and the `<matched-message>` fencing for voice-triggered watch fires are both thoughtful additions beyond the invited shape.

Four polish items for a follow-up (or the cleanup PR), none blocking:
1. `docs/audio-transcription-spec.md` and the `describeErrorBody` docstring still describe the *earlier* error handling ("first 200 chars of the body", "parsed for `detail.message`") — the code correctly does neither; update the prose to match its better self.
2. Neutralize `@channel`/`@here`/`@all` in the transcript echo for Mattermost — it's the one deterministic attacker-text→bot-post echo, and Mattermost pings on literal text where Slack doesn't. Cheap insurance.
3. Strip or escape a literal `</matched-message>` in `runner.ts` before fencing the matched text — defense-in-depth; the approval posture already limits it.
4. (Stray trailing blank line in `src/watches/runner.test.ts`.)

The equal-footing argument for not fencing the transcript in the prompt (it's the sender's own message, exactly as trusted as their typed text) is correct and well-articulated in the spec. Thanks for another one built right — this ships in 1.34.0 together with `!usage` once that gate lands.

---

Issue #528, opened 2026-09-02
Offer / discussion: voice mode — talk to the agent in a call, through the channel


kaza:
Hi Anne — this one is a question, not a PR. We built something and want to know whether it belongs anywhere near upstream, or stays a fork thing.

**What it is:** a browser page where you sign in with Slack, pick a task channel and talk. A voice model (Gemini Live today) acts as a front desk: it posts what you said into the channel *as you*, waits for the bot's reply, and reads it back. Claude never learns voice exists; it just sees a message in a channel. Two people can be on the same channel's call, and a Slack call card in the channel shows who is on it.

Code: https://github.com/kaza/claude-threads-upstream/tree/local/voice-desk/voice — a separate Bun service in `voice/`, ~2.5k lines with tests, no changes to the daemon except #527. Spec with the design and the review trail: `docs/voice-desk-spec.md` on that branch.

**Design points you may have opinions on:**

- Slack stays the bus. Any voice front end is a *writer into the channel*, never a change to the session model. That's why it needs #527 and nothing else.
- It posts as the person, with a user token they granted, so attribution and your allow-lists still hold.
- Replies are read once the streamed post has been quiet for ~8 s; there's no completion signal to read, so this is a heuristic. An "end of turn" marker from the daemon (a reaction, a metadata field, anything an integration can see) would make voice, and probably other integrations, a lot cleaner. Is that something you'd take?

**What we'd like to know:** is a `voice/` directory something you'd want upstream at all (opt-in, separate service, own env)? Or would you rather see it stay a sibling project that just needs #527? Either answer is fine; we run it regardless.

P.S. We'd have loved to put GPT-Live's voice on this, but it's a ChatGPT feature, not an API, so we went with Gemini Live. The front desk is one module; when OpenAI's side fits, it's a swap, not a rewrite. Same for the next step we're eyeing: dialling the agent from a phone.

anneschuth:
This is a genuinely impressive build, and the design instinct behind it — "Slack stays the bus, a voice front end is just a writer into the channel" — is exactly why #527 was worth merging. The daemon not knowing voice exists is the right invariant.

On your two questions:

[...]

PR #547, opened 2026-09-05
Turn marker: the daemon says when a turn is over (#528)

kaza:
The end-of-turn marker from #528, in the shape proposed there.

```yaml
turnMarker: metadata          # reaction | metadata | off (default)
# turnMarkerEmoji: checkered_flag   # reaction only; the default
```

After the `result` event's flush, the turn's last reply post is marked:

| `turnMarker` | What happens | Who sees it |
|---|---|---|
| `metadata` | the post gets Slack message metadata `event_type: claude_threads_turn_complete`, `event_payload: { session, turn, ok }` | integrations reading history with `include_all_metadata=true`; invisible in the UI. Slack only (refused on Mattermost at config time) |
| `reaction` | the bot reacts on the post with `turnMarkerEmoji` | everyone. Slack and Mattermost |
| `off` | nothing | |

`ok` is false when the turn ended with an error. A turn with no reply post marks nothing. A marker failure is logged and never touches the reply.

**How, and one design note.** The result flush op now carries the turn's outcome (`resultOk`). The marker is one dedicated write *after* the final flush — `chat.update` re-sending the post's current text with metadata, or `reactions.add` — rather than piggybacked on the final edit: splits, task-post reuse, empty flushes and failed writes all move "the final edit", so piggybacking marks the wrong post in exactly the cases that matter. One extra API call per turn is the price. Along the way the manager now awaits a scheduled flush that is still writing before running the result flush; that overlap existed before and was harmless until something depended on which post is last. Platform `createPost` / `updatePost` gain an optional `{ metadata }` (Mattermost accepts and drops it); `isErrorResultEvent` is exported from the CLI module.

Spec with the plan- and code-review trail: `docs/turn-marker-spec.md`; `docs/CONFIGURATION.md` documents the fields. 14 new tests; the affected suites are green.

Reader side, for the record: [claude-threads-voice-desk](https://github.com/kaza/claude-threads-voice-desk) now honours the marker and keeps its quiet-period rule as the fallback for a daemon without it. The `turn_waiting` question on #528 stands separately.


anneschuth (changes requested):
Verified independently on a worktree of 521880c: `tsc` and eslint clean, **3337 pass / 0 fail**, and the red-check is genuine — neutering `markTurnComplete` fails 4 tests, including the soft-flush-overlap one.

What I checked specifically, and what holds:

- **The DM call site is wired.** `src/index.ts:751` passes `turnMarker` through with the parent's values. That's the bug class that hit #529, #534 and #535 in one day; good to see it caught here without being asked.
- **`resolveTurnMarker` is properly defensive**: `metadata` refused on Mattermost at config time with a message that names the fix, emoji names validated whatever the mode, unknown values rejected with the field path.
- **`checkered_flag` collides with nothing** — it's not in `CANCEL_EMOJIS` or any control set, so a marker can't be mistaken for a user action.
- **The `flushInFlight` await earns its place.** Without it the marker lands on the wrong post, and your design note about not piggybacking on the final edit is right for exactly the reasons you give: splits and task-post reuse both move "the last post".

Two things before merge, neither large:

**1. `this.turn = 0` sits in dead code.** It's in `MessageManager.reset()`, whose only non-test caller is `dispose()`. The `!cd` / worktree respawn path calls `clearClaudeSessionState()`, which doesn't touch it. This is the same unreachable-reset finding as #534 — same method, and worth fixing once for both. The impact here is mild (a counter that keeps climbing across a `!cd` is arguably what you want), so the ask isn't "reset it" but "decide which it is and make the code say so". If climbing is correct, drop the line and note why; if not, wire it where #534's fix goes.

**2. `turn` isn't persisted, so a resumed session restarts at 1.** After a bot restart, `{session, turn}` repeats pairs it already emitted. voice-desk only asks "is this turn over", so it doesn't care — but the spec currently reads as though `turn` is monotonic per session, and the next integration will believe it. Either persist it beside the other session state or say plainly in `docs/turn-marker-spec.md` that `turn` is per-process and only `session` is durable. The doc line is fine by me.

Non-blocking: the `already_reacted` swallow is only on the reaction path, which is right, but a repeated `metadata` write on the same post (retry after a partial failure) is idempotent anyway — no action, just confirming I looked.

With those two, this merges. It's the cleanest of your open PRs and I'd like it in the next release.

kaza:
Both in.

**1. `this.turn = 0` in dead code.** Decided: the counter **climbs**, and the code now says so. `!cd`, a worktree switch and `!permissions interactive` replace the CLI process, not the conversation — restarting the count would re-issue `{session, turn}` pairs a reader has already seen, which is worse than a counter that keeps going. The line is gone, the declaration explains why, and a test pins it through `clearClaudeSessionState()` (the respawn path's actual call).

You were right that this is the same finding as #534. That one now has `MessageManager.clearTurnState()`, called by `restartClaudeSession` on every respawn — resume or not. If both land, `clearTurnState()` is the natural seam for anything that *should* reset per respawn; the turn counter deliberately isn't one of those things, so it stays out.

**2. `turn` not persisted.** Took the doc line, as offered. New section **`turn` is not durable** in the spec: `turn` orders within one run and repeats after a restart, `session` is the durable identity, readers dedupe by post id. The decisions table now records both this and the respawn choice, so the next integration reads it rather than inferring monotonicity.

**One more, from Gemini:** `handleFlushOp` awaited `flushInFlight` and *then* cancelled the scheduled flush. `await` yields the event loop, so a timer armed while an earlier flush was still running fires during the wait and starts a second flush — and cancelling afterwards is a no-op on a timer that already ran. That reopens exactly the overlap the await exists to close, and the marker can land on a post the late flush then supersedes. Cancel first, then await. The regression test drives the real sequence (arm, fire, arm again mid-flight, result) and asserts the two never overlap; verified red against the old order.

Thank you for the `already_reacted` note — I would rather hear that something was looked at and found fine than wonder.

3421 tests pass; typecheck and eslint clean. Reviewed by Codex and Gemini. **CodeRabbit did not run** — its free CLI tier hit its limit partway through this batch, so two reviewers here rather than three.

kaza:
Addendum — CodeRabbit ran after all (our CLI seat was misconfigured earlier, which is why my last comment said two reviewers). It found one more, and it was worth having:

**`this.turn++` sat after `await executeFlush()`**, so a rejected flush reused the number on the next turn. That was the one inconsistent case: a turn with no reply post already increments and emits nothing, so the counter tracks turns that *happened*, not turns that were successfully marked — and the spec I added says a gap is what tells the reader a turn went missing. Reusing the number hid exactly that.

Incremented before the flush, marker still written after it. No change on the success path. The spec now states the rule outright rather than leaving it to be inferred from the no-post example.

3422 tests pass; typecheck and eslint clean. Three reviewers on this one now: Codex, Gemini and CodeRabbit.

anneschuth (changes requested):
Thanks for the fast turnaround on the two asks and for the two extra fixes you pulled in from Gemini and CodeRabbit; all four are in and each has a red test.

Both asks from my first review are resolved:

- **`this.turn = 0` in dead code.** Gone from `reset()`, replaced by a comment that points at the field's declaration (`src/operations/message-manager.ts:557-565`), which now states that the counter climbs across a respawn and why. The test at `message-manager.test.ts:462` drives `clearClaudeSessionState()`, which is what the `!cd` and worktree paths call, and expects turn 3 after it. Climbing is the right call.
- **`turn` not persisted.** The new section "`turn` is not durable" in `docs/turn-marker-spec.md` says that `turn` restarts after a bot restart, that `session` is the durable identity, and that readers dedupe by post id. The decisions table records both this and the respawn choice. That is the doc line I asked for.

Two things before this merges.

**1. The `flushInFlight` guard covers timer flushes only; a `tool_complete` flush can still overlap the result flush, and the marker then precedes a late edit of the post it marks.** (`src/operations/message-manager.ts:462-482`, `:517-529`)

`scheduleFlush` records its `executeFlush` in `flushInFlight`, but `handleFlushOp` does not record its own. Every `user` event with a tool result emits a `tool_complete` FlushOp, and that flush runs from a `handleEvent` call the bot never awaits: `SessionManager.handleEvent` (`src/session/manager.ts:663`) fires `messageManager.handleEvent(event)` and returns, and `cli.ts:817-836` emits every line of one stdout chunk synchronously. So a `user`, `assistant`, `result` sequence starts three concurrent chains, and the result flush's `await this.flushInFlight` sees `null` while the tool-result update is still on the wire.

I reproduced it through the real code path (no stubbing of `executeFlush`, only a slow `updatePost` on the mock platform): an existing reply post, a tool result whose `updatePost` takes 60 ms, then the final text and result. Timeline:

```
start upd#1 "...`ls`\n\n↳ ✓"          tool_complete flush
start upd#2 "...↳ ✓\n\nDone."          result flush
done  upd#2
start upd#3 MARKER
done  upd#3 MARKER
done  upd#1 "...`ls`\n\n↳ ✓"          lands after the marker; "Done." is gone
```

The post is marked complete, then edited back to the text without the answer, and `currentPostContent` follows suit. In `createPost` form the same overlap produces two posts. The window is not academic: the Slack client honors `Retry-After` on a 429 (`src/platform/slack/client.ts:323-326`, default 5 s), so one rate-limited tool-result update stays in flight for seconds while Claude's final text and the result arrive. The content loss itself predates this PR; what this PR adds is a promise in the design note ("a soft flush still writing is awaited before the result flush, so the marker lands on the one post") that holds for half the flushes.

The fix is the same shape as what you did for the timer: record the `executeFlush` in `handleFlushOp` in `flushInFlight` as well (and await it in `flush()`), so any flush waits for the one before it. A test that drives `user`, `assistant`, `result` without awaiting between them, with a delayed `updatePost`, fails on the current branch and passes with that change.

**2. Rebase.** The branch conflicts with `main` in `src/index.ts`: your `resolveTurnMarker` import and #563's `resolvePlatformMcpPosture` import sit on the same line of the import list. Trivial to resolve, but GitHub reports the PR as unmergeable until it is.

Not blocking:

- **Relation to #534.** Both PRs add fields and a tail to `handleFlushOp` in `message-manager.ts`, so whichever lands second gets a four-hunk conflict there plus one in the test file; all are adjacent additions that resolve by keeping both sides. Two things for that merge. #534's `afterResultFlush` posts the tool details after the result flush, so `markTurnComplete` should run after it, keeping the marker the last write of the turn. And #534 names its method `clearTurnState()` while this PR's field is `turn`; a reader will assume the first clears the second. A comment in `clearTurnState()` saying it does not, or renaming the field to `completedTurns`, avoids the next confused review.
- **Interrupts.** `!escape` and the pause reaction send SIGINT; the CLI then emits an `error_during_execution` result (`real-cli-captures/interrupt.jsonl`), so the partial reply is marked with `ok: false` and the turn counts. That matches the spec and is the behavior I want.

What I verified on a worktree of e50c3a6: `tsc`, eslint and knip clean, 3340 pass / 0 fail. The DM call site still passes `turnMarker` through. `checkered_flag` collides with no control emoji, and both platform clients drop the bot's own reaction events before routing (`mattermost/client.ts:652`, `slack/client.ts:694`), so the marker reaction never reaches an executor. The `turn++` before `executeFlush` and the cancel-before-await order both do what the commits say.

anneschuth:
Status, so this is not waiting in the dark: I merged main into this branch and pushed (`33f7142`) since it had gone stale, and re-ran your red-green while I was there. Neutering `markTurnComplete` fails 6 tests in the turn-marker suite and restoring it passes 3478. That is a proper regression test.

My review round is still the thing holding this, not the conflict. It is on my list.

Three of yours went out in **1.36.0** today: `!usage` (#544), the reconnect-exhausted fix (#531) and `bugReports: false` (#575). Thank you for those, and for the correction you posted on #574 before anyone acted on it.

Also worth knowing: `Session Questions > decision bridge on option reaction` is currently flaky on the Mattermost job, failing at ~21.7s on unrelated PRs and main. If you see it red here it is probably not your change. Tracking in #584.

anneschuth:
I pushed the two asks from my second review plus main (`7e3b4d7`). Read it as a proposal and push back where you disagree.

**The overlap was real, and I want to be honest about how close I came to shipping a test that proved nothing.** My first two attempts at a regression test passed *without* the fix. The reason is worth knowing: `transformUser` only emits the `tool_complete` FlushOp for a tool whose start was recorded when its `tool_use` block was rendered (`transformer.ts:279`). My test fired a bare `tool_result` for a `tool_use_id` nothing had ever rendered, so no second flush existed and there was nothing to race. The write log was two writes, trivially ordered, green for entirely the wrong reason.

With a `tool_use` block in the preceding `assistant` event it reproduces through `handleEvent` alone:

```
ORDER: ["start 1","start 2","done 1","done 2", ...]
```

Write 2 begins while write 1 is still running, and write 1 lands last. Two concurrent `updatePost` calls on one post, completing out of order.

**The cause is narrower than "flushes can overlap".** Only `scheduleFlush` recorded into `flushInFlight`, so the waiter in `handleFlushOp` had nothing to wait on when the other writer was another `handleFlushOp` — and `SessionManager.handleEvent` does not await its handling, so a `tool_complete` flush and the `result` flush that follows it are exactly that pair. Every flush now goes through one `runTrackedFlush` helper, including `flush()`, which also waits for an in-flight write before starting its own.

I verified the red-green myself rather than trusting it: reverting just the tracked call in `handleFlushOp` fails exactly the new test and nothing else (3551 pass / 1 fail). The test now also asserts a minimum write count, so it can never go vacuous again the way mine did.

**`v: 1` is in the payload** (`{ v, session, turn, ok }`), with a test pinning both the version and the exact field set, so adding a field to a published format becomes a deliberate act rather than a diff nobody notices. Your reader ignores unknown fields already, so this is non-breaking for voice-desk today.

**And a stability statement in `CONFIGURATION.md`**, which is the thing I actually care about here. This marker leaves the process and is parsed by code we do not control; the moment someone builds on it we are committed. The paragraph says what is frozen (`event_type`, fields only added), what `v` means, that `session` is stable across restarts and respawns, and — the part that needed correcting — that **`turn` restarts at 1 after a bot restart and is an ordering hint within one run, never a unique key**. The spec said a gap always means a lost turn; that is only true for a reader that has watched an unbroken sequence, not one that reconnects mid-session. Readers that need exactly-once should dedupe by post id. Also a one-line note that each marked turn costs one extra API call, which operators running many concurrent sessions should see before they turn it on rather than after.

3552 pass, lint, typecheck and knip clean.

---

**Two things I am not changing, but want to put to you.**

`turn` is currently unread by the only consumer, and it carries the weakest guarantee of the three fields. Now that `v` exists you can drop it later without drama, but if you would rather not freeze it at all, say so and I will take it out before this lands. Either choice is defensible; I lean towards keeping it now that the doc is honest about what it means.

And `docs/turn-marker-spec.md` still reads as a design record rather than a contract — "Lessons learned: (none yet)" and per-reviewer attributions. Fine as a PR artifact, wrong as the document an integrator reads in two years. Worth moving the review trail into a "Design history" section at the bottom.

Approving once CI is green. Thanks for the patience on this one; three rounds is a lot, and the thing that made it worth it was your own design instinct to write the marker separately instead of piggybacking it on the reply.

anneschuth (approved):
Green. Items A through D from my review are in (`7e3b4d7`), and I verified the red-green independently: reverting just the tracked flush call fails exactly the new overlap test and nothing else.

The two open questions in my comment above — whether to keep `turn` now that `v` exists, and trimming the spec into a contract document — are yours to answer; neither blocks the merge. Leaving the merge to Anne.

Issue #533, opened 2026-09-02
Question: what is the intended pattern for keeping long-running agent work alive across a lost event?


kaza:
Not a bug report — a design question, and I'd rather ask than build something in a fork that turns out to be the wrong shape.

## The failure

You set an agent going on a long piece of work and leave it overnight. In the morning nothing has happened. Not because the work was hard, but because *one event went missing* six hours ago: an API blip, a dropped connection, a session that ended, a daemon restart during a deploy. The agent is sitting there waiting for something that is never going to arrive, and there is no signal at all — no error, no timeout, no post. You find out by discovering that nothing happened.

The cost is asymmetric in an unpleasant way. The fault lasts a second; the loss is the whole night.

#498, #499 and #500 are three concrete instances (PRs for all three just filed). But fixing individual causes does not answer the general question, because the next cause will be different — and some of them, like the Anthropic API dropping a connection, are not yours to fix at all.

## The question

**Is a stalled long-running session something claude-threads should notice, or is that deliberately out of scope?**

Both answers seem defensible:

- **Out of scope.** The bot is a bridge. A session that stops making progress is Claude's business, or the operator's, and building liveness detection into the bridge means guessing what "stuck" means — which is unknowable in general. Waiting 40 minutes for a long build is healthy; waiting 40 minutes for a message that will never come is not, and from outside they look identical.

- **In scope.** The bridge is the only component that can see "this session has produced no output for N minutes and nothing is pending", and it already owns session lifecycle, timeouts and persistence. Nobody else is positioned to notice.

## What exists today, as far as I can tell

- **`!routine`** — the closest thing. Time-triggered, persisted, and it survives both a session ending and a daemon restart, which is more than a `/loop`-style in-session mechanism can claim. But it fires `runUnattendedSession`, so it starts a *fresh* run rather than resuming the stalled one — a heartbeat that wakes a stranger. Useful only if the work's state lives outside the session.
- **`!watch`** — event-triggered, but on a triggering *message*, so it does not help with the absence of one.
- **Channel memory / distillation** — durable state exists, though it is oriented at conversational context rather than "what was I in the middle of".

So a workaround exists: keep a goal/state file in the worktree, and set a routine that reads it and either continues or reports. That works. It is also entirely convention — nothing in the tool suggests it, and every user has to invent it after losing a night first.

[...]

PR #549, opened 2026-09-05
Stop reporting a stalled mid-turn session as idle

### LINKED ISSUE #548: Idle reaper reports a stalled mid-turn session as 'timed out after N minutes of inactivity'
Split out from #533, where @kaza found this while pulling apart a proposed stall detector. It's a bug on main today, independent of that feature.

`cleanupIdleSessions` (`src/session/lifecycle.ts:2234`) decides a session is idle from `lastActivityAt` alone:

```ts
const idleMs = now - session.lastActivityAt.getTime();
if (idleMs > timeoutMs) { … "Session timed out after N minutes of inactivity" … }
```

There is no `isProcessing` check. So a session that is **mid-turn and stalled** — an API drop, a wedged CLI, a tool that never returns — gets reaped on the same path as a genuinely idle one, and the thread is told:

> **Session timed out** after N minutes of inactivity

which is false. It was not inactive. It was working, or waiting for something that never arrived, and the message sends the user looking in the wrong place.

Note that a long-running tool is not a false positive to be tolerated here but the *normal* case: events bracket a tool call (`tool_use` in the assistant event, `tool_result` in the user event) with nothing in between, so a ten-minute `Bash` is ten minutes of `lastActivityAt` not moving.

**Fix:** the reaper should distinguish the two. At minimum, when `isProcessing` is true the message should say so rather than claiming inactivity. Whether a stalled turn should be reaped at all, and after how long, is the larger question in #533 — this issue is only about the reaper not lying.

Related: #533 (the detector design), and the `openTurnAt` persistence work proposed there.

anneschuth:
Fixes #548, found by @kaza while taking apart the stall-detector spec in #533.

## The bug

`cleanupIdleSessions` decides idleness from `lastActivityAt` alone:

```ts
const idleMs = now - session.lastActivityAt.getTime();
if (idleMs > timeoutMs) { … "Session timed out after N minutes of inactivity" … }
```

There is no check on whether a turn is still open. So a session that is mid-turn and has simply stopped producing events gets reaped on the idle path and the thread is told it was inactive. It wasn't. It was working, or waiting for something that never arrived, and the message sends the user looking in the wrong place.

## Scope, which is narrower than the issue first suggested

`resetSessionActivity` runs on **every** Claude event, so a normally streaming turn keeps the clock warm and never reaches the reaper. The lie only surfaces when a turn goes silent:

- a wedged CLI or a dropped API connection (the #533 cases)
- one long-running tool call — events bracket a tool (`tool_use` … `tool_result`) with nothing in between, so a ten-minute `Bash` is ten minutes of a still clock

That third one makes this reachable in ordinary use, not just in failure.

## The change

Both the timeout and the pre-timeout warning now distinguish the two cases, keyed on `isProcessing` (set true when a message is sent, false on `result` — a clean invariant with exactly one setter each).

| | Between turns (unchanged) | Turn open, gone silent | Turn blocked on a human |
|---|---|---|---|
| Warning | `Session idle - will timeout in ~N minutes without activity` | `No output for a while - a turn is still running; will stop in ~N minutes if nothing arrives` | `Session still waiting for a reply - will stop in ~N minutes without one` |
| Timeout | `Session timed out after N minutes of inactivity` | `Session stopped responding - no output for N minutes while a turn was still running` | `Session still waiting for a reply - no answer for N minutes, so the turn was stopped` |

The third column came out of review. `isProcessing` stays true while a turn waits on a plan approval or a question, and the decision bridge waits an hour against a 30-minute default idle timeout — so "user gets a prompt, goes to lunch" routinely produced a post claiming the session stopped responding when the bot was waiting on *them*.

An ordinary MCP permission prompt is owned by the MCP child and is not visible to the bot process, so that case still reads as silence. Noted in the code; #533 carries the bridge line that would close it.

The log line gets the same split. Behavior is otherwise identical: same reaping, same `killSession(unpersist=false)`, same resume affordance. This only stops the bot from asserting something false.

Deliberately **not** in scope: whether a stalled turn should be reaped at all, and after how long. That's the #533 design discussion, and it needs `openTurnAt` persistence to answer properly. This is just the message.

## The regression this nearly shipped

Worth recording, since it was caught by review rather than by CI. The #491 bot-to-bot loop guard in `message-handler.ts` matches status posts by literal phrase:

```ts
new RegExp(`^⏱️\s+${BOLD}Session (?:timed out|idle)${BOLD}`, 'u')
```

New wording escapes it. Two claude-threads bots sharing a channel, one stalls, posts "Session stopped responding" — previously dropped as machine output, now routed as an ordinary message on the other bot. That is the unbounded loop #491 exists to prevent.

Every variant is now registered with the guard, with a comment on both sides pointing at the other.

A security pass then found the opposite failure in my first fix: the new phrases read far more naturally as human sentences than "Session timed out" ever did, and the guard's bold markers are optional with no trailing anchor. So `⏱️ No output for a while - is the build stuck?` was silently dropped, ack reaction and all, even when it @-mentioned the bot. Each pattern is now anchored on the clause the reaper actually emits, not the opening words.

Verified both directions against the real exported `isClaudeThreadsStatusPost`: all 6 emitted strings match on both formatters (Mattermost `**x**` and Slack `*x*`), and 4 plausible human phrasings pass through. Both sets are in the test table.

Also checked and clean: no attacker-controlled content reaches these strings (every interpolation is `Math.round` of a duration derived from `new Date()`), and the added patterns are linear under 800k chars of adversarial input.

Out of scope, filed as #551: the pre-existing authorization-refusal pattern on the line above backtracks quadratically (3.7 s at 80k chars, on the message hot path). One-character fix, but it isn't this PR's.

## Testing

9 new tests through the real `cleanupIdleSessions` (open turn, between turns, blocked on a decision — timeout and warning paths), asserting on the posted text, plus 10 added to the guard's `test.each` table (6 bot variants that must be caught, 4 human phrasings that must not).

Red-green verified three times: reverting the message logic fails exactly the two behavior-change tests (controls stay green), disabling the pending-decision branch fails its test, and reverting the guard regex fails 6 guard tests.

`bun test src/` 3377 pass / 0 fail, `tsc --noEmit` and eslint clean.

## Note for #529

@kaza — #529 gates these same call sites. Our hunks are adjacent but don't overlap (you wrap the `post()` calls, I changed the message strings above them), so this should merge cleanly either order. Flagging it so it isn't a surprise if git needs a nudge.

https://claude.ai/code/session_01USdFSTsyxgBgwvyp1UU491

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Fixed
-- **A session that stalls mid-turn is no longer reported as idle** (#548, found by @kaza while reviewing #533). The idle reaper decides from `lastActivityAt` alone, with no check on whether a turn is open. That clock is fed by every Claude event, so a streaming turn keeps it warm — but a turn that goes silent (a wedged CLI, a dropped API connection, one long-running tool call) was reaped with *"Session timed out after N minutes of inactivity"*, which is false and sends the user looking in the wrong place. The timeout and pre-timeout warning now distinguish the two cases: an open turn reports lost output, a genuinely idle session reads exactly as before.
+- **A session that stalls mid-turn is no longer reported as idle** (#548, found by @kaza while reviewing #533). The idle reaper decides from `lastActivityAt` alone, with no check on whether a turn is open. That clock is fed by every Claude event, so a streaming turn keeps it warm — but a turn that goes silent (a wedged CLI, a dropped API connection, one long-running tool call) was reaped with *"Session timed out after N minutes of inactivity"*, which is false and sends the user looking in the wrong place. The timeout and pre-timeout warning now distinguish three cases: a turn blocked on a plan approval or question says the bot is waiting on *you*, a turn that went silent reports lost output, and a genuinely idle session reads exactly as before. All variants are registered with the #491 bot-to-bot loop guard, so one bot's status post still can never read as a request to another.
 
 ## [1.33.1] - 2026-09-05
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- **A session that stalls mid-turn is no longer reported as idle** (#548, found by @kaza while reviewing #533). The idle reaper decides from `lastActivityAt` alone, with no check on whether a turn is open. That clock is fed by every Claude event, so a streaming turn keeps it warm — but a turn that goes silent (a wedged CLI, a dropped API connection, one long-running tool call) was reaped with *"Session timed out after N minutes of inactivity"*, which is false and sends the user looking in the wrong place. The timeout and pre-timeout warning now distinguish the two cases: an open turn reports lost output, a genuinely idle session reads exactly as before.
+
 ## [1.33.1] - 2026-09-05
 
 ### Fixed
--- a/src/message-handler.test.ts
+++ b/src/message-handler.test.ts
@@ -2858,6 +2858,14 @@ describe('isClaudeThreadsStatusPost (#491)', () => {
     ['⚠️ *Too busy* - 5 sessions active. Please try again later.'],
     ['⏱️ **Session timed out** after 30 minutes of inactivity'],
     ['⏱️ *Session idle* - will timeout in ~5 minutes without activity'],
+    // #548 variants: a stalled or decision-blocked turn reports differently
+    // from a genuinely idle one, and every variant must stay invisible.
+    ['⏱️ **Session stopped responding** - no output for 31 minutes while a turn was still running'],
+    ['⏱️ *Session stopped responding* - no output for 31 minutes while a turn was still running'],
+    ['⏱️ **Session still waiting** for a reply - no answer for 31 minutes, so the turn was stopped'],
+    ['⏱️ **No output for a while** - a turn is still running; will stop in ~4 minutes if nothing arrives'],
+    ['⏱️ *No output for a while* - a turn is still running; will stop in ~4 minutes if nothing arrives'],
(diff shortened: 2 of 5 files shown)


anneschuth (commented):
Independent verification pass (own-repo PR, so a comment rather than an approval — merging on the strength of it):

- **`isProcessing` invariant confirmed**: exactly one true-setter (message send) and one false-setter (`result`), with correct initializers (start = true, resume = false). The interrupt edge is closed — the real-CLI capture shows `!escape` emits a `result`, so the flag resets. Abnormal exit can't produce a wrong message (the exit handler removes the session before the reaper sees it).
- **Pending-decision detection is real state**, not a heuristic — it reads the QuestionApprovalExecutor's pending maps, which existing tests cover being nulled on resolution.
- **The #491 guard holds in both directions, beyond the PR's own table**: we reconstructed all 6 emitted strings through both real formatters — 12/12 dialect variants caught — and 6 human phrasings (your 4 plus 2 extra probes) pass through. Adversarial 800k-char input completes in ~10ms; linear as claimed.
- **Message-only claim confirmed byte-for-byte**: reaping, `killSession` args, persistence, and lifecycle-post logic untouched.
- **All three red-checks reproduced exactly** (guard revert → 6 fail/controls green; logic revert → 3 fail/controls green).
- **#551 confirmed real and worth prompt attention**: the pre-existing authorization-refusal pattern measures cleanly quadratic (2.2s at 80k chars) on the pre-routing hot path, and the one-char lazy quantifier fix measures 0.1ms. Correctly kept out of this PR.

Two accuracy nits, non-blocking: the body says "9 new reaper tests" — it's 5 (plus the 10 guard rows), and the guard table covers the Slack `*` dialect for only 2 of the 3 new phrasings (we verified the third matches too; a table row would pin it). One narrow observation for the #533 work: a `!cd` respawn doesn't reset `isProcessing`, so a session respawned mid-turn could carry a stale flag into the reaper — message-wording-only and pre-existing in spirit, noting it for completeness.

tsc/lint clean, 3377/3377, clean merge. Merging.

---

PR #553, opened 2026-09-06
Stop a long message from stalling message handling

### LINKED ISSUE #551: Quadratic backtracking in the authorization-refusal status pattern blocks the event loop
Found during the security pass on #549. **Pre-existing, not introduced there**, so filing separately.

The first pattern in `STATUS_POST_PATTERNS` (`src/message-handler.ts:56`) backtracks quadratically:

```js
/^⚠️\s+\S+ is not authorized\b/u
```

When ` is not authorized` never matches, the engine retries the `\s+`/`\S+` split at every position. Measured on `'⚠️ ' + 'a'.repeat(n)`:

| n | time |
|---|---|
| 10,000 | 61 ms |
| 20,000 | 232 ms |
| 40,000 | 930 ms |
| 80,000 | 3,717 ms |

Clean quadratic — 4x the time for 2x the input.

**Why it matters:** `isClaudeThreadsStatusPost` runs on **every inbound message**, on the single-threaded event loop, before routing. Mattermost's default post limit is 16k, so a message of `⚠️` followed by one long token costs ~150 ms of fully blocked event loop. Platforms or configs allowing larger posts degrade fast, and it needs no special privileges — any channel member can send it.

**Fix is one character:**

```js
/^⚠️\s+\S+? is not authorized\b/u
```

Measured at n=80,000: **3,717 ms → 0 ms.** A non-backtracking character class would work too.

Worth a regression test that asserts the match completes under a time bound, since the failure mode is latency rather than a wrong answer.

The other patterns in the list are linear — I stress-tested the `Session (?:timed out|idle|…)` alternation and the #549 additions to 800k chars of adversarial input and they all stay under 5 ms. This is the only one.

anneschuth:
Fixes #551, found during the security pass on #549 and filed separately because it predates that PR.

## The problem

`isClaudeThreadsStatusPost` runs on **every inbound message**, on the event loop, before the ack receipt and before any routing. Its first pattern was:

```js
/^⚠️\s+\S+ is not authorized\b/u
```

For a message that is *not* a refusal, ` is not authorized` never matches, so the engine retries the `\s+`/`\S+` split at every position before giving up. Measured on `'⚠️ ' + 'a'.repeat(n)`:

| n | greedy | lazy |
|---|---|---|
| 20,000 | 235 ms | 0.1 ms |
| 40,000 | 926 ms | 0.1 ms |
| 80,000 | 3,708 ms | 0.2 ms |

Clean quadratic. Mattermost's default post limit is 16k, so ~150 ms of fully blocked event loop per message — and it needs no privileges, just one long token from any channel member.

## The change

One character: `\S+` → `\S+?`.

The addressee is always a single whitespace-free token (`formatCode(username)`, an `@name`, or a raw `<@U…>`), so lazy matches exactly the same set. I verified that directly rather than assuming: **zero divergences** between the two forms across the matching cases (all three addressee shapes, extra whitespace) and the non-matching ones (the existing `lets ordinary messages through` table plus `⚠️ is not authorized` with no addressee).

I also stress-tested **every other pattern** in the list at 200k characters of `a`, `*`, `_`, and mixed filler against each of the five emoji prefixes. All stay under 3 ms. This was the only affected one.

## Testing

One regression test asserting the guard stays linear on long non-matching input. The bound is deliberately loose (250 ms for three 80k-char messages, against ~0 ms fixed and seconds broken) so it fails on real backtracking rather than on a slow CI runner.

Red-green verified: restoring the greedy quantifier fails it at **7,395 ms**, while every existing matching test stays green — so the fix changes performance, not behavior.

`bun test src/` 3396 pass / 0 fail, `tsc --noEmit` and eslint clean.

https://claude.ai/code/session_01USdFSTsyxgBgwvyp1UU491

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Fixed
-- **A long message no longer stalls message handling** (#551). The bot-to-bot status-post guard runs on every inbound message, on the event loop, before routing — and its authorization-refusal pattern used a greedy `\S+`, so any message that was *not* a refusal made the engine retry the split at every position before failing. Measured quadratic: ~150 ms at Mattermost's 16k post limit, 3.7 s at 80k, triggerable by any channel member with one long token. The quantifier is now lazy, which matches exactly the same strings (the addressee is a single whitespace-free token) in well under a millisecond.
+- **A long message no longer stalls message handling** (#551). The bot-to-bot status-post guard runs on every inbound message, on the event loop, before routing — and its authorization-refusal pattern used a greedy `\S+`, so any message that was *not* a refusal made the engine retry the split at every position before failing. Measured quadratic: ~150 ms at Mattermost's 16k post limit, 3.7 s at 80k, triggerable by any channel member with one long token. The quantifier is now lazy, which matches exactly the same strings (the addressee is a single whitespace-free token) in well under a millisecond. Measured under JSC, the engine Bun uses; V8 optimizes the greedy form away, so the fault is invisible when re-measured under Node.
 
 ## [1.34.0] - 2026-09-05
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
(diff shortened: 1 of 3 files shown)


anneschuth:
Independent review pass: no correctness findings. Two things it added that were worth folding in.

**The blowup is engine-specific.** It reproduced the quadratic behavior under JSC exactly as measured (235 / 923 / 3,706 / 14,819 ms at 20k / 40k / 80k / 160k), but under V8 both forms are sub-millisecond — V8 optimizes the greedy form away. We ship on Bun, so the original numbers were from the engine that matters, but the comment didn't say so and anyone re-measuring under Node would reasonably conclude there was nothing here. Now stated in both the code comment and the CHANGELOG.

**The regression test now has an explicit 30s bound.** In the broken state it takes ~7.4s, which also blew bun's 5s default, so a future regression would have reported as a runner timeout stacked on the assertion. It now fails cleanly as `Expected: < 250, Received: 7375`.

For the record, what the review verified independently rather than taking on trust: 200,000 randomized fuzz cases over an alphabet seeded with the pattern's own tokens, zero divergences between the greedy and lazy forms; the single-token addressee premise checked against all three emission sites (`message-handler.ts:469`, `:553`, `reaction-router.ts:191`) rather than assumed; adversarial inputs designed to defeat the lazy form specifically; all nine patterns re-stressed at 200k chars; and a sweep of the other inbound-path regexes (the `on branch\s+(\S+)` pair has the same shape but a literal prefix prunes the search, so it stays under 0.5 ms).

PR #555, opened 2026-09-06
fix(mattermost): probe quiet sockets with ping before declaring them dead

Jadefalkner:
## Problem

A channel with no traffic produces no WebSocket frames, which the heartbeat could not tell apart from a dead socket: after `HEARTBEAT_TIMEOUT_MS` of silence it forced a reconnect — every 90–150 s in a quiet channel, logged as `Connection dead (no activity for 90s)`. Any bot sitting in a low-traffic Mattermost channel reconnects in a loop all day.

## Fix

`BasePlatformClient` gains `sendHeartbeatProbe()`, called by the heartbeat tick once the socket has been silent for **half** the interval. The Mattermost client sends the `ping` action; the server's `pong` reply flows through `onmessage` and counts as activity, so only a socket that really stopped answering trips the timeout.

Half the interval, not the full one: the first tick lands a few ms short of `INTERVAL` and would skip the probe, and the next tick would then trip `TIMEOUT` (= 2 × `INTERVAL`) by the same few ms.

Slack is covered by #532 (protocol pings counted as activity); this is the Mattermost counterpart. Refs #498.

## Verification

- Production bot in a quiet private channel: 220 s without a reconnect (before: a reconnect every 90–150 s).
- New test in `src/platform/mattermost/client.test.ts`: a quiet socket is probed with `ping` and not declared dead.
- `tsc --noEmit`, eslint, platform suites clean.

### DIFF
--- a/src/platform/base-client.ts
+++ b/src/platform/base-client.ts
@@ -369,6 +369,13 @@ export abstract class BasePlatformClient extends EventEmitter implements Platfor
   }
 
   /**
+   * Send a request the server answers immediately (platform-specific). The
+   * reply must flow through the normal message path so it updates
+   * lastMessageAt. Default: no probe (heartbeat relies on organic traffic).
+   */
+  protected sendHeartbeatProbe(): void {}
+
+  /**
    * Start heartbeat monitoring to detect dead connections.
    * If no activity is detected for HEARTBEAT_TIMEOUT_MS, forces a reconnect.
    */
@@ -389,6 +396,16 @@ export abstract class BasePlatformClient extends EventEmitter implements Platfor
         return;
       }
 
+      // A quiet channel produces no traffic at all, which is indistinguishable
+      // from a dead socket. Ask the server for something cheap; its reply
+      // counts as activity, so only a socket that really stopped answering
+      // trips the timeout above.
+      // Half the interval, not the full one: the first tick lands a few ms
+      // short of INTERVAL and would skip the probe, and the next tick then
+      // trips TIMEOUT (= 2 × INTERVAL) by the same few ms.
+      if (silentFor >= this.HEARTBEAT_INTERVAL_MS / 2) {
+        this.sendHeartbeatProbe();
+      }
       wsLogger.debug(`Heartbeat check (last activity ${Math.round(silentFor / 1000)}s ago)`);
     }, this.HEARTBEAT_INTERVAL_MS);
   }
--- a/src/platform/mattermost/client.test.ts
+++ b/src/platform/mattermost/client.test.ts
@@ -363,13 +363,17 @@ describe('heartbeat probe', () => {
       ws: unknown; HEARTBEAT_INTERVAL_MS: number; HEARTBEAT_TIMEOUT_MS: number;
       startHeartbeat(): void; stopHeartbeat(): void; scheduleReconnect(): void; updateLastMessageTime(): void;
     };
-    Object.defineProperty(c, 'HEARTBEAT_INTERVAL_MS', { value: 15 });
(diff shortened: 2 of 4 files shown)


anneschuth (changes requested):
Thanks for this, and for measuring it on a production channel first.

Verified rather than just read:

- Sent `{action: "ping", seq: Date.now()}` over an authenticated socket to a Mattermost 11.7.8 server. The reply is `{"status":"OK","seq_reply":…,"data":{"text":"pong",…}}`, about 10 ms later, and it lands in `onmessage`, which calls `updateLastMessageTime()` before parsing. So the reply counts as activity exactly as described.
- The new test is red without the fix (with `base-client.ts` reverted to main it fails on `sent.length`) and green with it.
- The approach is the right one for the reason from #498: Node 22+/undici never surfaces server pings to JS, so the #532 route (counting them) cannot work there, while a client-initiated probe only uses the standard send path. Keeping the link as "Refs" is correct too, since Slack on Node 22+ has no probe yet and #498 stays open for that.

One change I'd like before merging: the test timing. Interval 15 ms, timeout 60 ms, real timers. The timeout check runs before the probe, so one event-loop stall of more than 60 ms between two ticks calls `scheduleReconnect` and the test fails on `reconnects`. Loaded CI runners stall that long. Interval 25 / timeout 250 / wait 500 keeps every assertion meaningful (the timeout can still trip inside the wait) with ten times the slack.

Nit: the `countPingsAsActivity` comment in `src/utils/websocket.ts` calls client-initiated pings "the eventual answer" for undici. Now that `sendHeartbeatProbe` exists, point it there.

Jadefalkner:
Done in 14175df: interval 25 / timeout 250 / wait 500 (ran the test three times in a row locally), and the undici note in websocket.ts now points at sendHeartbeatProbe.

anneschuth:
Still keen to land this: it closes the Mattermost half of #498 and the only open item is the test timing (interval 25 / timeout 250 / wait 500, or similar). If you don't get to it by Wednesday I'll push that tweak to the branch myself and merge; say so if you'd rather do it.

Jadefalkner:
Already on the branch: 14175df (interval 25 / timeout 250 / wait 500, undici note points at sendHeartbeatProbe) has been the PR head since Sunday, CI is green across the matrix and the branch merges cleanly onto current main. Nothing left on my side, feel free to merge.

anneschuth (approved):
Both asks are in, and I re-verified rather than took them on trust.

- Test timing is `interval 25 / timeout 250 / wait 500` (`src/platform/mattermost/client.test.ts:371-376`), with the CI-slack reasoning written down above it.
- The `countPingsAsActivity` note now points at `BasePlatformClient.sendHeartbeatProbe` (`src/utils/websocket.ts:41-43`).

On a worktree of 14175df: 3396 pass / 0 fail, lint, typecheck and knip clean. The new test is genuinely red — deleting the probe block at `base-client.ts:406-408` fails it on `sent.length` — and it survived 8 consecutive runs on the new timings, which was the whole point of the retune.

Two things I checked myself because a heartbeat change is easy to get subtly wrong:

- **Dead sockets are still detected.** A socket that accepts sends but never replies produced 9 probes and exactly one `scheduleReconnect`. The pong refreshes `lastMessageAt` because `onmessage` calls `updateLastMessageTime()` before the `JSON.parse`, so a quiet-but-live socket survives while a dead one still trips the timeout.
- **The half-interval guard is the load-bearing detail.** With `TIMEOUT = 2 × INTERVAL`, a full-interval guard would let the first tick land a few ms short, skip the probe, and have the second tick trip the timeout by those same few ms. The comment says exactly this, which is why I am not asking for a change.

Thanks for measuring it on a production channel first, and for the patience on the timing round.

PR #556, opened 2026-09-06
fix: set the session shutdown flag before the first await; resolve chmod via PATH in the test helper

Jadefalkner:
Two small, independent fixes; happy to split if you prefer.

## Shutdown ordering

`setShuttingDown()` ran after the 50 ms `await` that lets React paint the shutdown state. A service manager that signals the whole process group (systemd `KillMode=control-group`) delivers SIGTERM to the Claude children at the same instant as to the daemon. Their exit events arrived during that await, the flag was still unset, and lifecycle counted each exit as a resume failure. Three restarts in a row dropped the session (`Exceeded 3 resume failures`).

The flag is now set before the first `await`; everything else in the handler is unchanged.

## Test helper on NixOS

`setMode()` shelled out to `/bin/chmod`, which NixOS does not have (only `/run/current-system/sw/bin/chmod`). It now resolves `chmod` via `PATH`; the three affected unit tests (`setMode` SUID/SGID, `validateOutboundPath`) pass on NixOS again.

## Verification

- Session survived three consecutive `systemctl restart` cycles under `KillMode=control-group` (before: dropped on the third).
- `bun run test` 3425 pass / 0 fail on NixOS, `tsc --noEmit` and eslint clean.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -980,15 +980,17 @@ async function startWithoutDaemon() {
     // Guard against multiple shutdown calls (SIGINT + SIGTERM)
     if (isShuttingDown) return;
     isShuttingDown = true;
+    // Set the shutdown flag before ANY await: a service manager that signals
+    // the whole process group (systemd KillMode=control-group) kills the
+    // agent children at the same instant, and their exit events must not be
+    // mistaken for resume failures while we are still yielding to React.
+    session.setShuttingDown();
 
     // Update status bar to show shutdown in progress
     ui.setShuttingDown();
 
     // Give React a moment to render the shutdown state
     await new Promise((resolve) => setTimeout(resolve, 50));
-
-    // Set shutdown flag FIRST to prevent race conditions with exit events
-    session.setShuttingDown();
 
     // Update sticky messages to show shutdown state
     await session.updateAllStickyMessages();
--- a/src/test-utils/chmod-portable.ts
+++ b/src/test-utils/chmod-portable.ts
@@ -41,14 +41,15 @@ export async function setMode(path: string, mode: FileMode): Promise<void> {
   }
 
   // Step 2: shell out as a last resort. Use the absolute octal form so we
-  // don't depend on the symbolic parser of /bin/chmod.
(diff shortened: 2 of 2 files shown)


anneschuth (approved):
Thanks, both fixes are correct and I'm merging as-is; no need to split.

**Shutdown ordering.** Correct, and the mechanism it relies on holds. `SessionManager.getContext()` builds a fresh `SessionState` per call and `handleExit` fetches one at exit time (`manager.ts:678`), so the moved `setShuttingDown()` is visible to the next exit event. The 50 ms `await` was the only window: the module-level `isShuttingDown` in `index.ts` guards re-entry and never reached lifecycle. I looked for other shutdown paths with the same shape and found none; `SessionManager.shutdown()` sets its flag synchronously and there is no other SIGTERM handler in `src/`.

Worth knowing: `docs/systemd/claude-threads.service` sets no `KillMode`, so systemd's default `control-group` applies and every user of the documented unit hits this on every restart. That makes it a fix for the default install, not an edge case.

Two things I'd still want:

1. A test. The comment that was already there ("Set shutdown flag FIRST") did not stop the ordering from regressing once. The closure in `index.ts` is not testable as written, but extracting the sequence into a function that takes `{ session, ui }` would allow a test that asserts `setShuttingDown` was called before the first await, for instance with a session mock whose `updateAllStickyMessages` records whether the flag was already set. I'd accept the three-line move without it, but this is the kind of ordering a comment doesn't protect.
2. The header comment of `chmod-portable.ts` (step 3 of the "This helper:" list) still says `/bin/chmod`.

Residual, not blocking: the SIGTERM handler and the child's exit are still both event-loop events, so the ordering is a race in principle. The signal pipe fires in microseconds and the child needs milliseconds to die, so in practice this covers it. If it ever bites again, the durable fix is in `handleExit`: an exit by signal (`code === null`) that the bot did not send should not count toward `MAX_RESUME_FAILURES`.

**chmod via PATH.** Fine. Test-only helper, `execFileSync` without a shell, and macOS still resolves `/bin/chmod`. Ran `chmod-portable.test.ts` and `path-validator.test.ts` on the branch: 34 pass.

PR #557, opened 2026-09-06
feat: CLAUDE_THREADS_HOME state root, one-process-per-root lock, per-instance upload dirs

Jadefalkner:
## What

Run two bots under one user — one per platform, or a second one to try a new version — without them sharing state.

```yaml
# second instance
CLAUDE_THREADS_HOME=/home/bot/.second-bot claude-threads --headless
```

- **`CLAUDE_THREADS_HOME`** (`src/utils/state-home.ts`): every state path that resolved via `homedir()` now goes through `stateHome()` — config, sessions, github-emails store, platform list stores, thread logs, audit, memory, worktrees, update state. Existing explicit overrides (`CLAUDE_THREADS_SESSIONS_PATH` etc.) keep precedence. The override is resolved to an absolute path. Unset = `$HOME`, so the default behaviour is unchanged.
- **One process per state root** (`src/utils/instance-lock.ts`): a second start against the same root fails with a hint to set `CLAUDE_THREADS_HOME` and exits with status **3**, which the daemon wrapper treats as terminal — restarting would only collide again. A stale lock from a crashed process is taken over.
- **Per-instance upload directories**: two instances can share platform and thread ids (DCM sessions use the platform id as thread id), and one bot's session cleanup deleted the other's uploads under `$TMPDIR/claude-threads-uploads`. With `CLAUDE_THREADS_HOME` set the directory name carries a hash of the state root. The default instance keeps its historical path, so sessions persisted before this change still find their uploads.
- Docs: `CLAUDE_THREADS_HOME` in the environment table of `docs/CONFIGURATION.md`.

## Why

Config, `sessions.json`, thread logs, audit, memory, worktrees and the update state all live under `$HOME`, and two daemons silently share and overwrite them. I run a Claude-backed and a second bot side by side under one user; this is what makes that possible.

## Design notes

- **The lock without `flock()`.** Node 20 (the engine floor) has no file locking, so the lock is built from the two exclusive filesystem primitives that exist: `link()` never overwrites (EEXIST = a lock is held; the pid is written to a private temp file first, so the lock file never exists without content), and `rename()` moves a given inode exactly once (of several claimants of the same stale lock, one gets it). A claimant compares the inode it moved with the one it inspected — pid and inode are read from the same open file — and puts a fresh lock back with `link()` if it moved the wrong one. **No process ever unlinks another's lock**, and nothing is left behind by a claimant that crashes midway.
- **Remaining theoretical window, documented in the code**: three starters against a stale lock, one of them linking between two adjacent syscalls of another, can leave one holder unlocked. Closing that needs `flock()`. The practical deployment is one service unit per state root, which never produces simultaneous starts.
- Lock rejection uses its own exit status because `--restart-on-error` (the default for headless runs) would otherwise restart the rejected instance every two seconds forever.

## Verification

- Unit tests: live pid refused, dead pid taken over, re-acquire by the holder, ownership check on release, relative `CLAUDE_THREADS_HOME` resolved, upload path a single segment (existing traversal test in `tests/unit`), legacy path kept for the default instance.
- Stress: 30 rounds of 8 simultaneous starts against a stale lock and 10 rounds without one — exactly one survivor every round, no stray files.
- Two bots (Claude and a second backend) running side by side under one user in production for a day; a resumed session found its uploads after the upgrade.
- `bun run test` clean, `tsc --noEmit` and eslint clean.


anneschuth (changes requested):
Thanks, this is a clean way to get two bots under one user, and the lock design note made it reviewable.

What I checked: every `homedir()` state path on main is now routed through `stateHome()` (the one left, `BUN_INSTALL` resolution in the installer, is not state). The self-respawn path and the daemon wrapper both spread `process.env`, so the override survives an update restart. The MCP child imports none of the touched stores. The ownership guard in `release()` is red-tested: with the pid check removed, "release does not remove a lock that a successor took over" fails. Typecheck, eslint and the three test files pass locally; CI including both integration suites is green; `merge-tree` against main after #556 is clean.

Two things before merging:

1. **`link()` failure other than EEXIST aborts startup, for every install.** `tryLink` rethrows anything but EEXIST, and the lock is unconditional, so a state root on a filesystem without hard links (SMB/CIFS mounts without unix extensions, exFAT, rclone/gcsfuse style FUSE mounts) turns an upgrade into a bot that no longer starts, with an EPERM from `linkSync` as the only explanation. That user never asked for a second instance. Catch the non-EEXIST case, log a warning that the instance lock is unavailable on this filesystem, and run unlocked. The strong lock stays where `link()` works, and the default instance can never regress.
2. **Under systemd the new exit code still loops.** `docs/systemd/claude-threads.service` has `Restart=on-failure` with `RestartSec=10`, and the wrapper now exits 3, which systemd treats as a failure. So a second unit against the same root restarts every ten seconds forever, which is the loop the exit code was introduced to stop. Add `RestartPreventExitStatus=3` to the unit file, next to the comment about the wrapper handling restarts.

Smaller:

- `mkdirSync(dirname(path), { recursive: true })` in `acquireInstanceLock` creates `<root>/.config/claude-threads` with the umask default when it is the first writer, where `config/index.ts`, `platform-list-store.ts`, `audit-log.ts` and `memory/store.ts` all pass `mode: 0o700`. Under the daemon path the config loader has usually run first, but a fresh `CLAUDE_THREADS_HOME` in headless mode is exactly the case where it has not. Pass the mode. (`session-store.ts` and `github-emails-store.ts` have the same omission on main; not this PR's problem.)
- `hasStateHomeOverride()` is true for `CLAUDE_THREADS_HOME=$HOME`, so an explicit no-op override moves the upload directory and sessions persisted before it no longer find their files. Comparing `stateHome()` with `resolve(homedir())` avoids that.
- The refusal message names the pid but not the file. After pid reuse (host reboot, lock left by a SIGKILLed process, pid handed to something unrelated and long-lived) the bot refuses to start and the message points at a process that is not claude-threads. Include the lock path and say to delete it if that pid is not a claude-threads process.
- The comment in the inode-mismatch branch starts with "ponytail:". I assume that is a dictation artifact.
- Both test files `delete process.env.CLAUDE_THREADS_HOME` in their cleanup instead of restoring the previous value. Harmless in CI, surprising for someone running the suite with the variable set.

Noted, no change needed: the interactive self-respawn spawns the replacement while the old process is still alive, and the replacement is only refused if it reaches `acquireInstanceLock()` before the old process's `exit` handler has unlinked the lock. The old process exits within milliseconds of `spawnReplacement` and the replacement spends far longer than that on module load and the `claude --version` probe before it gets there, so the margin is wide. If it ever needs closing, a short retry on a live holder inside `acquireInstanceLock` would do it without touching the respawn code.

Jadefalkner:
Done in f06612c:

- link() failing with anything but EEXIST logs a warning and the instance runs unlocked, so a state root on SMB/exFAT/FUSE keeps starting after the upgrade. The strong lock stays where link() works.
- docs/systemd/claude-threads.service: RestartPreventExitStatus=3.
- mkdir with mode 0o700; CLAUDE_THREADS_HOME=$HOME is no longer treated as an override (compared against the resolved home); the refusal names the lock file and says to delete it when the pid is not a claude-threads process; the marker comment is a plain TODO; both test files restore the previous CLAUDE_THREADS_HOME.

Agree on the respawn margin; noted the retry idea in the code comment for when it is needed.

anneschuth:
Status while this waits on me, so it is not sitting in silence: your #555 landed today and is out in **1.36.0**, along with @kaza's #500 fix and `!usage`.

This one still has my earlier review round open. It is on my list, not forgotten, and the branch is clean against main.

One heads-up that affects your CI runs rather than your code: `Session Questions > should deliver the answer through the decision bridge on option reaction` is flaky on the Mattermost integration job right now, failing at ~21.7s on unrelated PRs and on main itself. If you see it red, it is very likely not you. I am chasing the real cause in #584; my first fix there was wrong, so it is still open.

Thanks for measuring the heartbeat change on a production channel before sending it. That is what made #555 easy to trust.

anneschuth:
Reviewed properly this time, and I pushed two small things to the branch (`a396e20`) rather than asking you for another round: current main, the lock file at `0600`, and a CHANGELOG entry. Read them as a proposal.

**The thing I actually needed to know, I tested rather than reasoned about.** This repo's hard rule is that persisted data keeps working across upgrades, and moving the state root is exactly the change that breaks it. So I built a populated pre-upgrade `$HOME` — config, `sessions.json` with two live sessions, routines, memory, thread logs, worktree metadata — and ran your code against it.

All six resolved state paths are byte-identical to main, both sessions are found and resume, and uploads keep their historical location. `stateHome()` is `resolve(process.env.CLAUDE_THREADS_HOME || homedir())`, so unset really is the same expression main used inline. **No migration is needed, and I would refuse one if offered**: it could only add bugs to a path that provably does not move.

I also checked the coverage held after today's four merges: no `homedir()`-rooted state path remains in `src/`.

**Red-green.** Three separate neuters, each genuinely red: `stateHome()` ignoring the env var fails 7 of 9; `isAlive()` always false fails the live-holder refusal; `release()` without the pid guard fails the successor-takeover case. Restored, 9/9. The tests call the real functions.

**The lock.** The design is careful in the way that matters: no process ever unlinks another's lock, the ownership guard is red-tested, and the hard-link fallback means an SMB or exFAT user cannot be locked out. I killed a holder with `-9` and the successor took over cleanly with no leftover files. The refusal message names the pid, the state dir, the fix and the escape hatch, which is the difference between an actionable error and a support ticket.

I checked the self-respawn path too, since a restart racing itself for the lock would be a nasty way to find out: the parent releases in ~0.14 ms and the child needs ~35 ms just to load the state modules before it reaches `acquireInstanceLock()`, inside one process's control flow rather than across a supervisor gap. `RestartPreventExitStatus=3` in the unit file makes the refusal terminal, and exit 42 still restarts.

**What I changed.** `instance.lock` was written `0644` while every other state writer sets `0600` explicitly. Nothing leaks (0700 parent, bare pid), but it was the one file deviating from the convention, so it now matches.

And the CHANGELOG says plainly that **two bots sharing one `$HOME` will no longer both start**. That is the only user-visible break here, it is deliberate and correct — those two instances were overwriting each other's `sessions.json` and silently losing sessions — but nobody should discover it when their second bot refuses to boot after an upgrade.

**One non-blocking note.** Nine state paths are module-level constants that call `stateHome()` at import time, so setting the env var after import is a no-op. That is fine in production, where the environment is set before launch, and it is pre-existing structure rather than something you introduced. But it is a live footgun for anyone writing tests later, and a one-line comment in `state-home.ts` saying the value must be set before module load would save someone an hour.

3544 pass, lint, typecheck and knip clean. This is good work, and the lock design note made it reviewable. Approving once CI is green.

anneschuth (approved):
Green. The backward-compatibility question is settled by test rather than by argument: a populated pre-upgrade state directory with two live sessions upgrades with both intact, and all six state paths resolve byte-identically to main.

Two small things pushed as `a396e20`: the lock file at 0600 to match every other state writer, and a CHANGELOG entry that says plainly what breaks. Leaving the merge to Anne.

PR #558, opened 2026-09-06
feat: agent backend abstraction and OpenAI Codex as a second backend

Jadefalkner:
## What

A second agent backend: OpenAI Codex via `codex app-server`, selectable per bot.

```yaml
agentBackend: codex   # default: claude
```

Sessions are driven through an `AgentSession` interface with two implementations: the existing `ClaudeCli` (unchanged behaviour, still the default) and `CodexSession`, which runs `codex app-server` over stdio (JSON-RPC, one process per session). Each session persists the backend it was started with, so old persisted sessions resume as Claude.

## How it works

`CodexSession` translates Codex notifications into the Claude stream-json event shape the rest of the bot already consumes — `system/init` (Codex thread id as `session_id`), assistant text / thinking / `tool_use`, user `tool_result`, `result` with `modelUsage`, `system/error` — so transformer, events handler, message manager, thread logs, audit and memory need no backend branches. Codex-native things are mapped onto the existing UI:

- **Approvals** (command execution, file changes, permission requests, MCP tool calls of foreign servers) go through the reaction prompts; the bot's own MCP server is accepted without a prompt, like on Claude. ✅ on an action prompt approves the rest of the session (`acceptForSession`). Timeouts, late reactions and interrupts close the prompts.
- **Questions**: `item/tool/requestUserInput` renders as `AskUserQuestion`.
- **Mid-turn messages** steer the running turn (`turn/steer`), like a stdin line does on Claude.
- **Commands**: `!compact` → `thread/compact/start`, `!model` / `!effort` → next `turn/start`, `!context` / `!cost` → token usage. `!stop`, `!cd`, `!permissions` and worktrees work unchanged.
- **Files**: image attachments are sent as `localImage` input; generated images are uploaded into the thread under the outbound-file policy (`outboundFiles.enabled` / `maxBytes`, validator). `send_file`, `read_post`, agent actions etc. come from the bot's MCP server, which Codex loads as `claude-threads-mcp`.
- **Limits and auth**: usage limits raise the existing rate-limit event with the reset time; auth errors are shown in the thread and treated as permanent.
- **No hidden Claude usage**: title/branch suggestions run through `codex exec --ephemeral`; startup validates the Codex CLI (installed, logged in, protocol version pinned to 0.153.4) and refuses `claudeAccounts` together with `agentBackend: codex`.

Permission mapping: `default` → `approvalPolicy: untrusted`, `sandbox: workspace-write`; `auto` → `on-request`; `bypass` → `never`, `danger-full-access`.

## Scope note

The `AgentSession` abstraction (interface, factory, five spawn sites) and the Codex adapter are bundled because the abstraction has no second user without the adapter. If you would rather take the abstraction alone first — or not take a second backend at all — say so and I will split or close.

The app-server protocol is marked experimental upstream; the bindings are pinned to one CLI version and drift only warns at startup.

## Verification

- Unit tests for the JSON-RPC client and the session translation: approval serialization, timeouts and late answers, allow-all, interrupt epochs, question sets and stale answers, permission requests, MCP elicitations, item translation, image generation, exit codes, steering, spawn failures, rate limits.
- The existing suite is unchanged (`bun run test` clean, `tsc --noEmit`, eslint).
- Hardened through adversarial review rounds with a second model; every finding is covered by a regression test.
- Running in production as a personal Mattermost bot (Codex, ChatGPT plan) next to a Claude-backed bot under the same user: approvals, questions, image generation, resume across restarts, usage-limit and re-login paths all exercised live.

### DIFF
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -52,6 +52,17 @@ platforms:
 | `threadLogs` | Thread logging (see below) | enabled |
 | `stickyMessage` | Sticky message text customization (see below) | none |
 | `claudeAccounts` | Multi-account pool (see below) | single-account mode |
+| `agentBackend` | Agent that runs new sessions: `claude` (Claude Code CLI) or `codex` (OpenAI Codex via `codex app-server`, see below). Existing sessions keep the backend they were started with. | `claude` |
+
+### Agent Backend (`agentBackend`)
+
+`agentBackend: codex` runs every new session on OpenAI Codex instead of the Claude Code CLI. The bot talks to `codex app-server` over stdio (JSON-RPC, one process per session), tested against Codex 0.153.4; startup checks that `codex` is installed and logged in (`codex login`) and warns on a version drift.
+
+What carries over unchanged: approvals via reactions (including "approve for the rest of this session"), questions from the agent (`item/tool/requestUserInput` is rendered like `AskUserQuestion`), `!stop`, `!cd`, `!permissions`, worktrees, file and image attachments (sent as `localImage` input), file uploads and agent actions through the bot's MCP server (Codex loads it as `claude-threads-mcp`), thread logs, audit log, memory, session resume across bot restarts (`thread/resume`). Backend-specific commands: `!compact` (`thread/compact/start`), `!model <name>`, `!effort <level>`, `!context` / `!cost` (token usage). Codex-side usage limits are reported in the thread with the reset time.
+
+Not supported with `codex`: `claudeAccounts` (Claude-only; startup refuses the combination), Claude slash-command passthrough, Claude plugins. Title and branch suggestions use `codex exec --ephemeral`. Codex additionally loads the MCP servers from its own `~/.codex/config.toml`; keep `allowedUsers` in mind when those servers reach personal data.
+
+Permission mapping: `default` → Codex `approvalPolicy: untrusted`, `sandbox: workspace-write`; `auto` → `on-request`; `bypass` → `never`, `danger-full-access`.
 
 ### Resource Limits (`limits`)
 
--- a/src/agents/codex/app-server.test.ts
+++ b/src/agents/codex/app-server.test.ts
@@ -1,0 +1,20 @@
+import { describe, expect, test } from 'bun:test';
+import { CodexAppServer } from './app-server.js';
+import { createLogger } from '../../utils/logger.js';
+
+describe('CodexAppServer', () => {
+  test('a spawn failure reports the error and releases the running state', async () => {
+    const rpc = new CodexAppServer(createLogger('test'), '/nonexistent/codex-binary');
+    const errors: Error[] = [];
+    const exits: Array<number | null> = [];
+    rpc.onError = (e) => errors.push(e);
+    rpc.onExit = (c) => exits.push(c);
+    rpc.spawn('/tmp');
+    const rejected = expect(rpc.request('initialize', {})).rejects.toThrow(/failed to start/);
+    await Bun.sleep(30);
+    expect(errors).toHaveLength(1);
+    expect(rpc.isRunning()).toBe(false);
+    expect(exits).toHaveLength(1);
+    await rejected;
+  });
+});
--- a/src/agents/codex/app-server.ts
+++ b/src/agents/codex/app-server.ts
@@ -6,7 +6,8 @@
  * responses `id`+`result|error`, notifications `method` only, and server→client
  * requests (approvals) carry `id`+`method` and expect a response with that id.
  */
-import { spawn, type ChildProcessWithoutNullStreams } from 'child_process';
+import type { ChildProcessWithoutNullStreams } from 'child_process';
+import { crossSpawn } from '../../utils/spawn.js';
 import { createInterface } from 'readline';
 import type { Logger } from '../../utils/logger.js';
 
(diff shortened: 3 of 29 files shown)


anneschuth (changes requested):
Technical review first. Whether claude-threads takes on a second backend at all (as-is, abstraction only, or not) is a separate call that I'll answer in this thread after thinking it over; nothing below assumes an answer.

Thanks for the careful work here, in particular the review rounds you already folded in; the adapter reads well.

What I verified: `tsc`, eslint and knip are clean on the branch, the full suite passes (3433 tests), and two of the new tests go red when I break the code they cover (`sessionApproved` on allow-all; the identity-guarded `clear` in `handleApprovalResponse`). The Claude path is unchanged in every spot I traced: `buildPermissionArgs` is a pure extraction, the reserved approval slot and the guarded clear are no-ops for a single pending plan approval, the init `session_id` adoption never fires because the CLI echoes the id we pass, `runBackendCommand` returns false when the backend has no `runSlashCommand`, and the modelUsage fallback only kicks in when no model carries a cost. Persisted sessions without `agentBackend` resume as Claude. Merges clean on current main. I could not exercise the Codex side itself: the Codex install on this machine is broken, so everything below about the wire protocol takes your captures on trust.

Blocking:

1. **✅ allow-all is not audit-logged.** In `question-approval.ts` the 👍 branch calls `auditDecision(true)` and the 👎 branch `auditDecision(false)`; the new ✅ branch at line 458 calls neither. The one decision that silences every further prompt for the session is the one missing from the audit trail. It should log as an approval with an allow-all marker.

2. **Codex approvals use the plan-approval gate, not the permission-prompt gate.** The reaction router admits `session.sessionAllowedUsers` (which includes `!invite`d guests) plus the platform allowlist unless approvals mode is `owner`. The MCP permission prompt on the Claude path only accepts `ALLOWED_USERS`, i.e. the platform allowlist. On Codex, a guest who was invited to talk in a thread can approve command execution, file changes, network access and ✅ allow-all for the rest of the session. That is a different privilege model for the same emoji, and it is not mentioned in the PR or the docs. Either gate `action` approvals (at least ✅) the way the MCP server does, or document the difference and make it deliberate.

3. **The permanent-failure regex is too broad.** `/unauthori|log ?in|sign in|refresh token|access token|auth/i` on any `error` notification sets `permanentFailure`, and lifecycle then unpersists the session on its first failed resume ("Detected permanent failure, removing from persistence"). `auth` matches "author", "authorization header rejected", an OAuth error from a foreign MCP server listed in `~/.codex/config.toml`, none of which mean the bot user must run `codex login`. You already use `codexErrorInfo === 'usageLimitExceeded'` for rate limits; the auth case deserves the same structured check, with the message regex at most as a narrow fallback.

4. **Comments in German and private markers.** `src/agents/types.ts` (header and the `getLastStderr` doc), `src/agents/factory.ts`, "handoff Punkt 8" in `src/index.ts`, and two `ponytail:` markers in `src/agents/codex/session.ts`. The codebase is English, and "Umbaufolge Punkt 4" refers to a document nobody else has. Please translate and turn the markers into plain TODO comments that say what is missing.

Wanted before merge, not blocking on their own:

- `app-server.test.ts` covers only a spawn failure. The JSON-RPC framing (request id matching, server request → response with the same id, notification dispatch, error mapping, the EPIPE listener) is untested, and it is the layer a Codex release is most likely to break. Feeding lines into `handle()` directly would cover most of it without a process.
- There is no way to run the Codex path in CI. The integration mock speaks the Claude dialect and the protocol is pinned to one CLI version that ships weekly upstream. A recorded app-server transcript (the way `tests/integration/fixtures/real-cli-captures/` works for Claude) replayed through `CodexAppServer` would at least catch dialect drift before users do.
- One question I could not settle here: does the app-server write the per-thread `config.mcp_servers` block, including the `PLATFORM_TOKEN` env, into its rollout files under `~/.codex/sessions/` or its logs? On the Claude side the same env goes into a `0600` temp file for that reason. Please grep a rollout after a session and say what you found.
- `docs/CONFIGURATION.md` says memory carries over unchanged. Only the channel layer does (it rides in `developerInstructions`); Claude's native auto-memory redirect has no Codex counterpart. Also `CODEX_BIN` is new and undocumented; `CLAUDE_PATH` has a row in the env table.
- Every haiku one-shot (title, tags, distillation, routine and watch parsing, each watch confirm) becomes a `codex exec` at the user's default Codex model, 6 to 7 seconds each by your own measurement. Worth one sentence in the docs so operators can size the watch caps accordingly.

Nits:

- `approval_timeout` closes a prompt post with "the request expired" also after an interrupt or a finished turn.
- The terminal header shows "Claude unknown" on a Codex bot (`claudeVersion: 'unknown'`).
- `CLAUDE.md`'s source tables do not know about `src/agents/`; a short row set would help the next reader.

anneschuth:
Thanks again for the work here, and for the review rounds you already went through. I'm not going to take a second backend into claude-threads, so I'm closing this rather than have you fix the blocking items for a merge that isn't coming.

The reason is maintenance, not the code. Everything that makes the bot useful is built against the Claude CLI's contract: the permission-prompt tool and the decision bridge, the account pool, the memory redirect, and the stream-json dialect with captured fixtures and an integration mock that speaks it. A Codex adapter is a second such contract, against a protocol its own authors mark experimental, pinned to one CLI version that ships weekly, with no way to run it in CI. Every upstream release could break it, only users would notice, and that lands on me. The abstraction on its own doesn't change the calculation: without a second backend it has no user, so I'd rather not carry it either.

If you keep this in a fork, the review findings above still apply, the audit gap and the approval gate in particular. Should the app-server protocol stabilize and real demand for it show up here, I'm open to revisiting.

Jadefalkner:
Understood, and thanks for the thorough review regardless. I'll keep it in the fork and fold in the audit and approval-gate findings there. If the app-server protocol settles down I'll come back with numbers.

PR #564, opened 2026-09-06
1.35.0

anneschuth:
Minor release. Merging this publishes to npm via release.yml.

- `package.json` 1.34.2 to 1.35.0 (`package-lock.json` follows).
- CHANGELOG `[Unreleased]` promoted to `[1.35.0] - 2026-09-06` with #563: sessions only see the bot's MCP servers (`--strict-mcp-config` by default), new `mcpServers` and `strictMcpConfig` config keys, MCP server logging at session start.

The changelog entry is marked breaking because sessions that relied on inherited MCP servers lose them without an error; it ships as a minor on purpose, with the two ways back (`mcpServers`, `strictMcpConfig: false`) spelled out in the entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [1.35.0] - 2026-09-06
+
 ### Changed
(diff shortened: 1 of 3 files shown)


anneschuth:
Superseded: the strict-by-default change in #563 is being reworked before release (connectors-only exclusion). A fresh release PR follows once that lands.

PR #569, opened 2026-09-07
fix: clear all listeners when removeAllListeners() is called with no argument

R09722akaBennett:
## Summary

`TypedEventEmitter.removeAllListeners()` removes nothing when called without an
argument, so `MessageManager.dispose()` never drops the session's listeners and
every finished session leaks the closures it registered.

## Changes

- `src/operations/message-manager-events.ts`: forward a no-arg
  `removeAllListeners()` call to `EventEmitter` as a no-arg call, instead of
  passing an explicit `undefined`.

## Root cause

The override forwards the optional parameter unconditionally:

```ts
removeAllListeners(event?: MessageManagerEvent): this {
  return super.removeAllListeners(event);
}
```

`EventEmitter.prototype.removeAllListeners` branches on `arguments.length`. A
no-arg call on the subclass still reaches the base method with
`arguments.length === 1` and `type === undefined`, which is read as "remove
listeners for the event named `undefined`" -- so nothing is removed.

`MessageManager.dispose()` relies on this call, and its own comment describes
the leak it is meant to prevent:

```ts
// Without this the closures held by listeners keep session state alive
// after the session ends, leaking across many sessions.
this.events.removeAllListeners();
```

## Testing

Two existing tests already cover this behaviour and are currently failing on
`main`:

- `TypedEventEmitter > removeAllListeners > removes all listeners when no event specified`
- `MessageManager > Lifecycle > removes all event listeners on dispose`

Both pass with this change. Full unit suite goes from 3394 pass / 2 fail to
**3396 pass / 0 fail**. `bun run lint`, `tsc --noEmit` and `bun run build` all
succeed.

Reproduced identically on bun 1.4.2 and node 26, so this is not a
runtime-version quirk.

## Checklist

- [ ] Documentation updated (if needed) -- not applicable

### DIFF
--- a/src/operations/message-manager-events.ts
+++ b/src/operations/message-manager-events.ts
@@ -252,7 +252,13 @@ export class TypedEventEmitter extends EventEmitter {
    * @returns this (for chaining)
    */
   removeAllListeners(event?: MessageManagerEvent): this {
-    return super.removeAllListeners(event);
+    // Forward a no-arg call as a no-arg call. EventEmitter branches on
+    // arguments.length, so an explicit `undefined` reads as "remove listeners
+    // for the event named undefined" and clears nothing -- which silently left
+    // dispose() leaking every session's listeners.
+    return event === undefined
+      ? super.removeAllListeners()
+      : super.removeAllListeners(event);
   }
 
   /**


anneschuth (approved):
Thanks, good catch, and the diagnosis is exactly right.

What I checked: on Node 25 a subclass that forwards the optional argument leaves every listener in place after a no-arg call, and the fix clears them. The two tests you name pass on main under Bun 1.3.3, which is what CI pins, and fail under Bun 1.4.2, which matches Node here; with this branch they pass on 1.4.2 and the full suite is green there too. So the leak was real in the published build (it runs under Node) while CI could not see it. That is a CI problem of ours, not of this PR; the Bun pin is going to move.

Merging.

Issue #570, opened 2026-09-07
Design: a routine fire that resumes a paused thread instead of starting a fresh session (#533 item 4)


kaza:
Sketching answers to the three design questions you raised on #533 before writing any code, as asked. Also the piece #550 turns out to depend on.

## Why this shape

`!routine` already survives session end and daemon restart, which is what makes it the only real building block for long-running work. What it cannot do is *nudge the work that stalled* — it fires `runUnattendedSession`, which anchors a fresh root post and starts a new session. A heartbeat that wakes a stranger.

Adding a target to the existing scheduler is the whole feature:

```yaml
!routine add watchdog "check the migration and continue or report" --every 1h --target thread
```

`--target new` (default) is today's behaviour, unchanged.

### 1. What does the fire say?

**Not a bare "continue".** That is the automatic-continuation hazard we both agreed to hold on: a `tool_use` with no recorded `tool_result` can be re-issued, and the external effect may already have happened. A routine that says "continue" makes that rare ambiguity routine.

Instead the routine's own prompt is delivered into the existing conversation, framed so Claude knows where it is:

```
[Scheduled routine "watchdog" — fired on its schedule into this session,
not sent by a live user. The conversation above is your own earlier work.
Decide whether anything is still outstanding, then either carry on or
report where things stand.]

<routine prompt>
```

The difference from the current prefix is one clause: today it says *"post the result in this thread"* because the thread is new. Here the thread is the agent's own history, and saying so is what makes the nudge useful rather than confusing. The decision to continue stays Claude's, made with the history in front of it — which is exactly the judgement we did not want to automate.

### 2. What if the thread's session is active rather than paused?

**Skip, and say so in the log.** Not queue.

Only a *paused* session may be resumed, for the reason that also settles question 3: **the respawn is the only point at which posture can legally change.** A live session's MCP child already holds whatever permissions it was launched with; a boolean flipped afterwards revokes nothing.

The three cases:

| Session state | Fire does | Why |
|---|---|---|
| Paused (tombstoned `stopped` excluded) | Resume it and inject the prompt | The case the feature exists for |
| Active | Skip, log, no post | It is alive. A nudge would interleave with a live turn, and there is nothing to resume |
| No persisted session at all | Skip, log, no post | A routine that resumes must not quietly become a routine that starts — that is `--target new`, and the user chose otherwise |

[...]

Issue #550, opened 2026-09-05
DCM sessions cannot propose routines or watches — the suppression may have outlived its reason


kaza:
Splitting this out of #533, where you raised it:

> *On the DCM `propose_routine` refusal: that's the self-replication/injection gate being deliberately blunt, and your watchdog case is a fair example of it overreaching. The confirmation-card flow (human 👍 required) is the actual safety boundary; the DCM refusal predates trusting it fully. Worth its own small issue if it's blocking you.*

### What it blocks

An agent working alone in a task channel cannot propose its own watchdog. That is the exact case where one would help most: a long unattended job, in a channel-as-task setup, where nobody is watching. The pattern has to be set up in advance by a human who already knows to — which is the discipline nobody has at the moment they say "keep going overnight".

### Where it lives

```ts
// src/mcp/mcp-server.ts:1651
// DCM sessions can list but never create routines/watches — don't offer
// propose_* tools the bot can only refuse.
const noProposals = unattended || process.env[AGENT_FEATURES_ENV.DCM] === '1';
```

so `propose_routine` (`:1677`) and `propose_watch` (`:1695`) are simply not registered for a DCM session. `list_routines` / `list_watches` stay, as `agent-features-env.ts:28-31` describes.

### The part we could not verify

Both comments justify the suppression as avoiding a tool "the bot can only refuse" — the MCP side is hiding something it expects to be rejected downstream. **We could not find that downstream refusal.** Searching for a DCM gate in the routine/watch creation path, the confirmation-card flow, the command executor and the stores turned up nothing that rejects a `dcm:` thread; the only `isDcmThreadId` use nearby is reply threading (`message-manager.ts:373`).

Either it lives somewhere we did not look, or the suppression has outlived the refusal it was shielding — in which case removing it is a one-line change rather than a behaviour change. Worth knowing which before anyone writes a patch, and you will know immediately where we would be guessing.

### Why the safety argument still holds

The self-replication concern is real: an agent that can schedule its own future runs can, in principle, keep itself alive. But the boundary you named is the one doing the work — `propose_routine` **creates nothing**. It posts a confirmation card, and only a human 👍 on that card saves anything (`mcp-server.ts:118-120`). An agent that proposes a watchdog every hour is a nuisance in a thread, not a self-replicating process.

[...]

PR #571, opened 2026-09-07
resume: stop promising that interrupted work continued (#533 item 1)

kaza:
Item (1) from #533 — the one you said you'd take today, and #549 only covered the reaper message so it was still open.

## The lie

The resume notice said **"You can continue where you left off"** on every path, including the one where nothing continues. After a deploy, a thread whose turn was in flight is told it can carry on; `isProcessing` starts `false`, nothing is sent to the CLI, and the work is simply gone. The thread looks resumed and is idle.

Both branches shared the sentence, so both were wrong after a restart — and the *edit-an-existing-post* branch is the one a deploy actually takes, since `postShutdownMessages` sets `lifecyclePostId` on the way down.

## The fix

The wording now turns on **why** the resume happened.

| Trigger | Notice says |
|---|---|
| A person (message or reaction) | "Reconnected… You can continue where you left off." |
| Daemon boot | "Session resumed after bot restart (v1.35.0) — conversation history is intact, but anything that was still running when the bot stopped did not survive the restart. Send a message to pick it up." |
| Platform re-enabled | Same, worded for the platform being disabled rather than a restart |

**Three review rounds, and two of them found this PR making smaller versions of the same mistake:**

- Gemini: the edit branch still opened with *"Session resumed by @owner"* — on a boot resume that credits someone who wasn't there. Attributed only when a person asked, now.
- Codex: my first cut inferred "the daemon did this" from `resumedBy === undefined`. That's wrong. `manager.ts:1092` is `resumePausedSessionsForPlatform`, which fires when an **operator re-enables a platform from the UI toggle** — no username, no restart. Those threads would have been told their work didn't survive a restart that never happened. The reason is now passed explicitly (`ResumeTrigger`), not deduced. Fixing a message that guessed by guessing would have been a poor look.
- Codex also checked a comment I wrote claiming a person's resume is always followed by their message: `tryResumeFromReaction` returns without sending anything to Claude. The invitation is still right for that path — someone is present and about to type — but not for the reason I'd written down. Comment corrected.

## Scope

This does **not** detect that a turn was interrupted — that needs `openTurnAt`, item (2). It stops claiming otherwise. Nothing else changes: no new posts, no suppression, no lifecycle behaviour.

Four tests, three verified red by reverting their branch individually. 3533 tests pass; typecheck, eslint and `git diff --check` clean. Reviewed by Codex, Gemini and CodeRabbit (0 findings on the final diff).

---

Separately, I've opened #570 with the design sketch you asked for on item (4) — the three questions answered, plus how it subsumes the DCM half from #550.

### DIFF
--- a/src/session/lifecycle.test.ts
+++ b/src/session/lifecycle.test.ts
@@ -1337,6 +1337,91 @@ describe('authorization gate at sinks (#388)', () => {
     });
   });
 
+  describe('the resume notice tells the truth about what continued', () => {
+    // A resume with no `resumedBy` was nobody's request: the daemon restarted
+    // and restored the session. `isProcessing` starts false and nothing is
+    // sent to the CLI, so a turn that was in flight when the bot stopped is
+    // simply gone — and the notice used to say "you can continue where you
+    // left off" regardless (#533).
+    let prevClaudePath: string | undefined;
+    beforeEach(() => {
+      prevClaudePath = process.env.CLAUDE_PATH;
+      process.env.CLAUDE_PATH = '/bin/echo';
+    });
+    afterEach(() => {
+      if (prevClaudePath === undefined) delete process.env.CLAUDE_PATH;
+      else process.env.CLAUDE_PATH = prevClaudePath;
+    });
+
+    function bootState(overrides?: Record<string, unknown>) {
+      return {
+        threadId: 'thread-boot',
+        platformId: 'test-platform',
+        claudeSessionId: 'claude-session-1',
+        workingDir: process.cwd(),
+        startedBy: 'alice',
+        sessionAllowedUsers: ['alice'],
+        ...overrides,
+      };
+    }
+
+    function bootContext() {
+      const platform = createMockPlatform({
+        isUserAllowed: mock(() => true) as any,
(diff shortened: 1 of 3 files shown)


anneschuth (approved):
Thanks, and for passing the trigger instead of guessing it from a missing username; that was the right correction to take from the review rounds.

What I checked: both daemon call sites in `src/session/manager.ts` name their trigger (`boot` at the resume-all, `platform-enabled` in `resumePausedSessionsForPlatform`), the two person-initiated paths still carry `resumedBy`, and both post branches (edit the lifecycle post, create a fresh one) use the same outcome sentence. Typecheck clean, the lifecycle suite passes, and reverting the boot wording turns a test red. Merges clean on current main.

Two notes, neither blocking:

- Dropping the `'platform-enabled'` argument at the manager call site keeps every test green: the platform-enabled wording is tested through `resumeSession` directly, not through `resumePausedSessionsForPlatform`. A test that toggles a platform and reads the post would close that gap.
- The comment block above `outcome` says the same thing twice (the "What actually continues depends on WHY" paragraph and the "A person who asked for this resume" paragraph). Keep the second one.

Merging; this goes out in 1.35.1.

PR #575, opened 2026-09-08
config: bugReports: false — a switch that keeps bug reports on-premises (#574)

### LINKED ISSUE #574: Feature: a config switch to disable !bug, for deployments that cannot send data off-premises
## The problem

`!bug` is the one command that sends session data **off the operator's infrastructure**, and there is currently no way to turn it off short of forking.

| What leaves | Where it goes |
|---|---|
| Attached screenshots | `catbox.moe` — a public anonymous file host (`bug-report/handler.ts:31`), which the code describes as "permanent hosting" |
| Report body | `anneschuth/claude-threads`, hardcoded, public (`bug-report/handler.ts:190`) |
| Session context **plus the last 50 daemon log lines** | same public issue (`bug-report/handler.ts:365`) |

Redaction is regex plus `@redactpii/node` — best-effort by construction, which is the right design for a feature people opt into, and not something a compliance team will accept as a control.

Two details make it sharper than "a user might paste something sensitive":

- **`claudeCanExecute: true`** (`commands/registry.ts:334`). The agent can file a public issue itself, with no human typing anything.
- The 🐛 **error reaction** reaches the same path, so a user can trigger it without knowing they have.

None of this is wrong for the project's own users — a bug reporter that just works is a genuinely good feature, and the log lines are exactly what makes a report useful. It is simply disqualifying in a regulated environment.

## Why I'm raising it rather than forking

This came from a company I'm consulting for. They are heavily regulated, their security review found this path, and their options today are "don't deploy claude-threads" or "maintain a fork that deletes a feature". Both are worse for you than a config flag, and I doubt they're the only ones — "can we prove nothing leaves our network" is a standard question at that end of the market.

A fork is what we'd do otherwise, and I'd rather not: it's a permanent divergence over one boolean.

## Proposal

```yaml
bugReports: false     # default true — nothing changes unless you set it
```

Behaviour when `false`:

- `!bug` is **refused with a short explanation**, not silently dropped. An unknown command falls through to the dynamic passthrough and reaches Claude as an ordinary prompt, so "make it not exist" would send the report text to the model rather than nowhere.
- The 🐛 error reaction does nothing.
- Claude invoking `!bug` itself hits the same refusal.
- A report card already pending from **before** the switch was thrown cannot be approved — persisted state survives the restart that applies the config, and approving one would still file the issue. Denial still works, so the card can be cleared.
- `!bug` disappears from `!help`, since advertising a command that will be refused only generates support questions.

The gate goes in `reportBug`, which is the single funnel all four entry points pass through — gating the command table alone would leave the reaction and the agent path open.

**One deliberate difference from the sibling flags:** it fails closed. `resolveBooleanFeature` warns and falls back to the default on a malformed value, which is harmless for `auditLog` or `watches`. Here the default direction publishes data, so `bugReports: "flase"` in a config an operator wrote *in order to restrict the bot* must not quietly leave the path open.

## PR

Implemented and offered as a PR alongside this issue — default `true`, no behaviour change for existing users, seven tests, full suite green. Happy to change the shape (per-platform instead of global, different naming, silent instead of explicit refusal) or to close it if you'd rather this stayed a fork-level concern; I'd just rather ask than diverge quietly.

kaza:
Closes #574 if you want it; happy to close the PR instead if you'd rather this stayed a fork-level concern.

## What

```yaml
bugReports: false     # default true — nothing changes unless you set it
```

`!bug` is the only command that sends session data off the operator's infrastructure: screenshots to `catbox.moe` (public, anonymous, "permanent hosting" per the code's own comment), and a body carrying session context **plus the last 50 daemon log lines** to a hardcoded public repo, behind best-effort redaction. `claudeCanExecute: true` means the agent can file one with no human typing anything.

That's the right design for this project's users and disqualifying for a regulated deployment, where the alternative today is a fork that deletes a feature. #574 has the full argument.

## Where the gate goes

In `reportBug`, which is the single funnel for every route that creates a report:

| Entry point | Reaches `reportBug` |
|---|---|
| `!bug <text>` typed by a user | `commands/executor.ts:512` → `manager.ts:1548` |
| Claude invoking `!bug` itself | `operations/events/handler.ts:120` |
| 🐛 error reaction | `session/reaction-router.ts:292` |

Gating the command table alone would have left the other two open.

The approval callback is gated separately in the manager, because it reaches `gh issue create` **without** passing through `reportBug`. Approval becomes a denial: the card clears, nothing is sent, and an explicit denial still works. `!bug` also comes out of `!help`, since advertising a command that will be refused only generates support questions.

**Refusal is explicit, not silent.** An unknown command falls through to the dynamic passthrough and reaches Claude as an ordinary prompt — so "make it not exist" would send the report text to the model rather than nowhere.

**It fails closed, unlike its siblings.** `resolveBooleanFeature` warns and falls back to the default on a malformed value, which is harmless for `auditLog` or `watches`. Here the default direction publishes data, so `bugReports: "flase"` — in a config written to *restrict* the bot — disables the feature and says so. A bare `bugReports:` (YAML `null`) counts as malformed for the same reason: the operator wrote the key.

## What the review round changed

Three reviewers, and two findings were mine to own:

- **The tests didn't protect the thing that matters.** They passed no attachments and watched no upload, so **moving the gate below the catbox upload left all of them green** — the switch would still refuse, after the screenshots had already left. There's now a test with an attachment asserting `downloadFile` is never called, which is the earliest observable sign of data leaving the host. That mutation now fails three tests.
- **I justified the approval gate with something false.** I claimed a pending card survives the restart that applies the config. It doesn't — `MessageManager.serialize` carries only the task list and the context prompt. The gate stays because the callback is a real second door; the comment now says that instead. Corrected on #574 too.
- CodeRabbit caught `null` being treated as absent. Taken.
- Gemini flagged that Claude's system prompt might still advertise `!bug`. **Rejected** — `generateChatPlatformPrompt` filters the agent-executable list to `worktree` and `cd`; running the generator and grepping for "bug" returns nothing.

Codex audited every caller of `uploadImages`, `uploadImageToCatbox` and `createGitHubIssue` and confirmed no route reaches either when the flag is false, and that the gate precedes the CLI check, context collection, attachment download and upload.

## Scope

Default `true`, so no behaviour change for anyone who doesn't set it. The required `SessionConfig` field meant the compiler found all three call sites that needed updating. Ten tests. 3542 pass; typecheck, eslint and `git diff --check` clean.

Happy to reshape it — per-platform instead of global, different naming, silent refusal instead of explicit — or drop it. I'd just rather ask than diverge quietly over one boolean.


anneschuth (changes requested):
Thanks for bringing this as a PR instead of a fork, and for the correction on #574. The gate is in the right place and I could not find a route around it. Two things need to change before merge, both small.

**1. The approval-callback gate has no test that goes red (`src/session/manager.ts:489`).** I removed `&& this.bugReportsEnabled` from the `handleBugReportApproval` lambda and the full suite stays green (3542 pass). Removing the gate in `reportBug` fails three tests, so that half is protected; this half is not. I agree it is unreachable today, since `reportBug` refuses before `setPendingBugReport` and `MessageManager.serialize` never persists a pending card. That is exactly why it will be deleted by the next person who reads it as dead code. Either give it a test or move it. My preference is to move it: give `commands.handleBugReportApproval` a `ctx: SessionContext` parameter, check `ctx.config.bugReportsEnabled` there next to the existing `if (!pending) return`, and add a fourth case to `bug-report-gate.test.ts` with a pending report set on the mock message manager, asserting the card is cleared and `createGitHubIssue` is never reached. That keeps both doors gated in one file and both testable with the harness you already built. If you keep it in the manager, `manager.test.ts` constructs a `SessionManager` positionally and can reach `getContext().ops.handleBugReportApproval`.

**2. The CHANGELOG and the docs row still carry the claim you retracted.** `CHANGELOG.md:12` says the switch closes "a report already pending from before the switch", and `docs/CONFIGURATION.md:57` says "an already-pending report cannot be approved". You showed on #574 that a pending card does not survive the restart that applies the config, so neither sentence describes something a user can observe. Drop both; the comment in `manager.ts` already tells the true story.

**3. The CHANGELOG entry will conflict with #573.** I merged this branch onto #573 with `git merge-tree` and got a conflict in `CHANGELOG.md`: both insert directly under `## [Unreleased]`. Once 1.35.1 is on main, put the entry under a fresh `[Unreleased]` heading above `## [1.35.1]`. Not your doing, just a heads-up so the rebase is a one-liner.

Smaller points, no action required:

- `postError` still adds the 🐛 reaction to every error post when the flag is off (`src/operations/post-helpers/index.ts:125`, `src/operations/message-manager.ts:1047`). The bot invites a click it then refuses with the disabled notice. Harmless, and it needs a `ctx` that `postError` does not have, so I would leave it unless you see a cheap way.
- The `bugReportsEnabled: true,` line added to the three test contexts left the following `maxSessions` line at the wrong indentation (`events/handler.test.ts:132`, `monitor/handler.test.ts:25`, `test-utils/mock-session-context.ts:53`).

What I verified on the branch: typecheck, eslint, knip and the unit suite are green. All three entry points reach `reportBug`: the typed command via `commands/executor.ts:510`, Claude's own `!bug` via `operations/events/handler.ts:120`, the 🐛 reaction via `session/reaction-router.ts:292`. `createGitHubIssue` has exactly one caller (`handleBugReportApproval`) and `uploadImages` exactly one (`reportBug`); the MCP server exposes no bug tool and the auto-update path never files a report. The system prompt generator filters the agent-executable list to `worktree` and `cd`, so Claude is not told about `!bug`, and the parser's `!bug` match still lands in the gated function if Claude emits one anyway. Config: absent key resolves to `true`, `null` and any non-boolean resolve to `false` with a warning, and the loader has no top-level key whitelist that would drop `bugReports`. The single `SessionManager` construction site in `src/index.ts` passes the resolved value; the integration harness falls back to the default `true`. The branch is based on current main and merges clean apart from the CHANGELOG note above.

kaza:
All four in, plus the optional one — the precedent made it cheap enough to take.

**1. The approval gate now goes red.** You're right that an untested gate on an unreachable path is a gate with a deletion date. Moved into `commands.handleBugReportApproval` as you suggested: it takes `ctx` and checks `bugReportsEnabled` next to the existing `if (!pending) return`, so both doors are gated in one file and both reachable from the harness that was already there. The new case sets a pending card, approves it with the flag off, and asserts the card is cleared and no issue URL ever reaches the post. I ran your mutation — removing that check now fails it.

Worth noting the two halves protect different things, which is why the split bothered me once you pointed it out: removing the `reportBug` gate fails three tests, and *moving it below the catbox upload* fails those same three. Removing the approval check failed nothing at all.

**2. The retracted claim is gone** from both `CHANGELOG.md` and `docs/CONFIGURATION.md`. Replaced with what the switch actually closes — the typed command, Claude invoking it, the 🐛 reaction, the approval that files the issue, and the `!help` listing.

**3. Merged 1.35.1** and moved the entry under a fresh `[Unreleased]` above it. The conflict is gone.

**4. Indentation fixed** in all three test contexts. Mine, from a scripted insertion.

**And the 🐛 reaction — I took it.** You were right that `postError` has no `ctx` and 25 call sites, but `configureAuditLog` already establishes the module-level configure pattern for exactly this shape, so `configureBugReports` follows it and `index.ts` calls it beside the existing one. Both error-post paths are covered (`post-helpers/index.ts` and `MessageManager.postError`), with a test that flips the switch and asserts the reaction appears and disappears.

It's a global, which I know is not free. Happy to drop it if you'd rather not carry one for this — the feature is complete without it, and it's a single commit to revert. My argument for keeping it: in the deployments this switch exists for, "the bot still shows you the button and then refuses" is the kind of detail a security review writes down.

3545 tests pass; typecheck, eslint and `git diff --check` clean. The two node-e2e failures I hit were the new suite from #578 needing `dist/` built first — green after `bun run build`, and unrelated to this branch.

anneschuth (approved):
All four are in, the optional one too, and I verified each rather than reading the diff. Merging.

The mutation is the one that mattered, and it now goes red: commenting out the `bugReportsEnabled` check at `src/operations/commands/handler.ts:1450` takes the suite from 3462 pass / 0 fail to 1 fail, on `refuses to file a report card that is already pending`. Restoring it goes green. Lint, typecheck and knip are clean.

The gate placement holds up under tracing. There are exactly two egress primitives, one caller each — `uploadImages` → catbox.moe and `createGitHubIssue` → `gh issue create` — and the gate at `:1345` sits above the `downloadFile` binding at `:1383`, not merely above the upload. So no attachment bytes are read off the platform, let alone sent. All four trigger paths (typed command, Claude's own `!bug`, the 🐛 reaction, and the approval callback) land behind one of the two checks. Config fails closed: a bare `bugReports:` is `false` with a warning, not `true`.

**One thing to fix in a follow-up, and it is not theoretical.** The pending-card test does not stub `createGitHubIssue`, so with the gate removed it shells out to the real `gh` CLI against this repo. My mutation run filed #583 ("T"/"B"), which I have closed. That also explains #581 — same title, same body, filed the day you ran the mutation yourself. So the test that proves the gate is load-bearing is a live-fire round: anyone who repeats the mutation, or breaks the gate in a future refactor, files a public issue from their test suite.

Not a defect in what ships here, and not a reason to hold the PR: the shipped gate is exactly right. But please stub `createGitHubIssue` in that case. I will close #581 as the same accident.

Two cosmetic leftovers, neither blocking:

- `bug-report-gate.test.ts:7` still carries the retracted "pending from before the switch was thrown" claim you removed from the CHANGELOG and docs. The in-test comment at `:99-102` gets it right.
- `getBugReportsEnabled()` (`manager.ts:1539`) has no production caller yet.

Thanks for the correction on #574 and for taking the optional fifth item — suppressing the 🐛 affordance when the feature is off is the difference between a switch and a trapdoor.

PR #584, opened 2026-09-10
Fix the flaky decision-bridge question test

anneschuth:
`Session Questions > should deliver the answer through the decision bridge on option reaction` has failed on four runs in a row today, always at ~21.7s, on unrelated PRs (#579, #580, #544) and on main itself. That is a race, not bad luck, and it was costing a rerun on every PR.

**The race.** The bot posts the question text first and adds its own 1️⃣–4️⃣ option reactions afterwards. The test reacted as soon as the text matched, so on a slow runner it could react before anything was listening. The answer is dropped, and the second `waitForPostMatching` then burns its full 20s timeout, which is exactly the ~21.7s signature.

**The fix.** Wait for the bot's own `one` reaction before adding ours. `waitForReaction` already polls for precisely this, so it is one line plus the import. No production code changes.

I cannot reproduce this locally (no Docker here), so the real verification is this PR's own integration run.

https://claude.ai/code/session_013bNsyE8EoWg6c2hStCGyJx

### DIFF
--- a/tests/integration/suites/session-questions.test.ts
+++ b/tests/integration/suites/session-questions.test.ts
@@ -16,6 +16,7 @@ import {
   waitForPostMatching,
   getThreadPosts,
   addReaction,
+  waitForReaction,
   getPlatformBotOptions,
   type TestSessionContext,
 } from '../helpers/session-helpers.js';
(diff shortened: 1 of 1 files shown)


anneschuth:
Closing in favor of #587, which fixes the actual cause.

My diagnosis here was wrong. I assumed the test reacted before the bot had added its option reactions, and that waiting for the bot's own 1️⃣ would close the gap. It did not: this branch still failed at 21702ms, the same signature.

The real problem is on the bot's side, not the test's. The post is registered in the session's reaction index only after `createInteractivePost` resolves, while the post has been live and reactable since before the first option emoji was sent. A reaction in that window is dropped silently, and for `AskUserQuestion`/`ExitPlanMode` that parks the decision bridge until `MCP_TOOL_TIMEOUT`.

So the test was not flaky, it was right. Worth noting that my fix here made the failure *more* likely, since waiting for the first option lands the reaction squarely in the remaining gap.

PR #587, opened 2026-09-10
Claim interactive posts before the option reactions are added

anneschuth:
Supersedes #584. What looked like a flaky integration test turned out to be a real bug, and the test was reporting it correctly.

## The bug

A question or plan-approval post is created first, and its 1️⃣/2️⃣ (or 👍/👎) options are added afterwards, one API round trip each (`base-client.ts:349-358`). The post is live and reactable for that entire window, but the bot only claimed it once `createInteractivePost` had resolved:

- `message-manager.ts:509` registered the post id in the session's reaction index after the await
- `question-approval.ts:250` / `:200` recorded `currentPostId` / `pendingApproval` at that same late moment

A reaction landing in between hits `registry.findByPost` with an unindexed id and is **discarded silently**, no retry and no fallback.

The consequence is worse than a missed click. On a modern CLI, `AskUserQuestion` and `ExitPlanMode` block on the MCP permission prompt, so the decision bridge stays parked and the session hangs until `MCP_TOOL_TIMEOUT`, an hour by default, on a decision the user already made. The faster someone answers, the more likely they lose the answer.

## The fix

`createInteractivePost` gains an optional `onPostCreated` callback that fires as soon as the post exists, before the first option emoji. Registration and the executor's pending-state claim both move into it. The shared `postInteractiveAndRegister` helper is fixed too, which covers nine other reaction-gated prompts: worktree, message approval, update, bug report, routine and watch confirmations.

The post-hoc assignments stay as a defensive fallback for a platform override that ignores the callback, but they are guarded by a `claimed` flag so they can never re-arm a decision the user has already consumed. Without that guard the fallback would re-open an answered question.

## Verification

Red-green both ways. Neutering the two callbacks fails exactly the two new cases in `question-approval.test.ts` and passes the other 8; restoring them passes all 10.

- Deterministic repro with an injected latency window: 0/6 before, 11/11 after
- The flaky suite in isolation: 10/10 after
- Full unit suite 3528 pass / 0 fail, typecheck, lint and knip clean

`executor-harness.ts` previously baked in the buggy ordering, so it was fixed to mirror the real sequence. Left alone deliberately: `subagent.ts:165` has the same pattern, but it only governs a minimize/expand toggle, so missing one leaves the post readable and re-clickable with no blocked bridge.

## Why it only showed on CI

The window is ~5 ms on an idle machine, which is why this never reproduced locally and why the integration test passed 5/5 in isolation. A loaded CI runner widens it enough to hit reliably.

https://claude.ai/code/session_013bNsyE8EoWg6c2hStCGyJx


anneschuth:
Note before anyone merges this: the integration job here **still fails the same test**, at 21741ms, with this fix in place.

That does not invalidate the fix. I verified the red-green independently: neutering both `onPostCreated` callbacks fails exactly the two new cases and passes the other 8; restoring them passes 10/10, with the full suite at 3528 pass and typecheck, lint and knip clean. The race is real, the reasoning about the blocked decision bridge holds, and it is worth fixing on its own merits.

It just is not the whole story for this particular test. The likely reason is an *earlier* gap than the one closed here: `waitForPostMatching` matches on the question text via the test's own API poll, so the test can react before the bot process knows the post exists at all, which is before `onPostCreated` has run. This fix closes the window between post-created and options-added; it does not close a window between post-created-on-the-server and post-created-known-to-the-bot.

If that turns out to be the case, the interesting question is whether the reaction router should fall back to a thread lookup when it sees a reaction on an unknown post id, rather than dropping the event outright. Dropping user input silently is the pattern behind both halves of this.

Investigating now. Holding the merge until the job is green rather than merging a fix whose own regression test is red on CI.

anneschuth:
Green, and the test that failed all day now passes in **1776ms** instead of timing out at 21.7s. That is the verdict I was holding the merge for.

My earlier comment guessed there was a gap before the one this PR originally closed. There is, and it is worse than the first:

`src/platform/mattermost/client.ts:342` sends `POST /posts`. The server stores the post, and starts accepting reactions on it, the moment it handles that request. The bot only learns the post id when the **response** gets back, so the post is reactable strictly before its id can be indexed. Caught on the wire with the first fix in place and no artificial latency:

```
[repro2 ...552] question visible to test: 64q3... — reacting NOW
[PROBE  ...558] createPost returned id=64q3... (request sent at ...541, +17ms)
```

The test saw the post 6ms before the bot knew it existed, and passed only because the reaction's own round trip happened to be slower. Widening the response lag failed it 100%, with the reaction dropped 205ms before the id landed.

**So there were two bugs stacked**, and both silently discarded input a user really gave:
1. the post id was indexed only after the option emojis were added (the original commit)
2. the post is reactable before its create response returns at all (the new one)

The second is fixed by announcing the create *before* issuing it and having a reaction on an unknown post wait up to 5s for that registration, rather than dropping an event the platform never re-delivers. The wait applies only while a create is genuinely outstanding, so a reaction on any unrelated post still returns immediately and there is no DoS surface. Waiters are keyed by post id, so a concurrent create on another thread can never resolve a reaction against the wrong session.

Verified independently rather than taken on trust: removing just the router wait fails the new router test and passes the other 18; restoring it passes 92/92 across router, registry and question-approval. Full suite 3535 pass, typecheck, lint and knip clean.

The integration test was not weakened. It still reacts immediately on the question text and still asserts the continuation post, so it proves the whole round trip. The entire fix is in the product.

Merging.

